PR #1, opened 2017-04-19
Updating the report on assignments and getting omage ranges from gv file

younes-elhachi:
First commit and pull request on github :-) .
I added more printing on the ring assignments as disscussed last time.
I Needed to "estimate" the measured data (omegra ranges) from gv file in order to compute the epected rings and peaks per ring.
Maybe it is better to just ask the user for the omega ranges!

p2
p1

### DIFF
--- a/ImageD11/indexing.py
+++ b/ImageD11/indexing.py
@@ -325,7 +325,9 @@ class indexer:
                         best=diff
         # Report on assignments
         ds=np.array(self.ds)
-        print "Ring     (  h,  k,  l) Mult  total indexed to_index  "
+        print "Ring     (  h,  k,  l) Mult  total indexed to_index  ubis  peaks_per_ubi"
+        expeectedubis = []
+        expectedpeaks =[]
         # try reverse order instead
         for j in range(len(dsr))[::-1]:
             ind = np.compress( np.equal(self.ra,j), np.arange(self.ra.shape[0]) )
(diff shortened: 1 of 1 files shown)


jonwright:
Hi Younes,

I merged it - seems good. There is a simulated dataset in 
ImageD11\test\ken_simul>python idx.py

Having 19 estimate grains is about right (there are 20).

Can you make a couple of changes in the printing? Where it estimates 
min_peaks this should be "min_pks" and the value is something like the 
sum of the last column. To reach a min_peaks value in the range 5-7 we 
would need to delete all the other rings or change the way we could 
min_pks...

Thanks!

Jon

===


Got wavelength from gv file of  0.153636
Got wedge from gv file of  0.0

Got 1 sets of omega values from gv file:
- Range 1 from  -60 to   60 ==> 120 degrees
Read your gv file containing (1805L, 3L)
...
setting: pars[cosine_tol] from 0.002 to 0.002
Maximum d-spacing considered 1.209699
Ring assignment array shape (1805L,)
Ring     (  h,  k,  l) Mult  total indexed to_index ubis  peaks_per_ubi
Ring 8   ( -2, -2, -4)   24    309       0      309 19     16
Ring 7   ( -4, -2,  0)   24    319       0      319 19     16
Ring 6   ( -3, -3, -1)   24    310       0      310 19     16
Ring 5   ( -4,  0,  0)    6     87       0       87 21      4
Ring 4   ( -2, -2, -2)    8    107       0      107 20      5
Ring 3   ( -3, -1, -1)   24    316       0      316 19     16
Ring 2   ( -2, -2,  0)   12    157       0      157 19      8
Ring 1   ( -2,  0,  0)    6     90       0       90 22      4
Ring 0   ( -1, -1, -1)    8    110       0      110 20      5
In case you use ring 1 and ring 5 for the indexing,
you might set the min_peaks to something between  5 and 7

younes-elhachi:
Hi Jon,

Ok I will improve the printing as by your suggestions.
For the value of min_pks, I summed only two values of the last column (for the two "best" rings picked up according to the column before last). So I will print also the min_pks value of the whole data (sum of all the last column).

May be, we should add a submenu for doing the assignment / trial / score / writeubis with only the proposed two rings (or different rings that the user indicate)!

Thank you for committing my slight changes!
Best
Younes

jonwright:
Hi Younes,

Some other improvements to think about:

- if the "indexer" object works with a columnfile instead of the gve 
file: it could more easily call refinegrains so that grid_index can be 
integrated in the gui. (This is quite a big change internally).

- we could produce gve files from the indexing menu for grainspotter - 
ask the user to select which rings to use for indexing in grainspotter. 
We can also pass the omega ranges to create the input file for 
grainspotter...

younes-elhachi:
Hi Jon,

I agree, the makemap / grid_index should be integrated in the gui.
I'm checking this to figure out how the indexer oject is linked to other scripts

PR #3, opened 2017-04-24
Producing GrainSpotter input file

younes-elhachi:
Hi Jon,

I added here the first functions to deal with GrainSpotter whenever it's necessary.


younes-elhachi:
A quick test script is also given in test/grainspotter/test_gspotter.py

younes-elhachi:
Hi Jon,

I was not sure about these functions being part of the indexer class or not. So thank you for your suggestions.
I will make a new module for driving grainspotter with better methods.

I was thinking about reading grainspotter output but I thought it was complicated to do!
Now that you propose the same idea, this means that it's feasible.

PR #56, opened 2019-07-03
Situation with installion on MacOSX

jonwright:


jonwright:
pip install should be starting to work from git master, so I close this, see #105

PR #93, opened 2020-03-31
Peaksearch from NXS files

jadball:
This is an altered peaksearcher.py script that reads files to search and omega motor position values from a Nexus file. Changed command line arguments accordingly.


jonwright:
Seems good. I will merge for now and just send in another pull request if you want to change something more, in sandbox things are pretty relaxed. For the peaksearcher in ImageD11src it is a bit more painful to test it (people usually complain at some point when I make a mistake).

PR #118, opened 2021-02-09
New feature: expert 2theta histogram to text file

smerkel:
Hi: a quick change to ImageD11 gui: export 2theta histograms to text file for further analysis

### DIFF
--- a/ImageD11src/tkGui/guitransformer.py
+++ b/ImageD11src/tkGui/guitransformer.py
@@ -52,6 +52,7 @@ class guitransformer:
               ( "Save parameters", 0, self.saveparameters),
 #              ( "Plot selected columns", 0, self.plotcols),
               ( "Plot tth histogram", 0, self.plothisto ),
+              ( "Export tth histogram", 0, self.savehisto ), 
               ( "Filter peaks based on tth histogram", 0, self.filterhisto ),
               ( "Compute g-vectors", 0, self.computegv),
               ( "Save g-vectors", 0, self.savegv),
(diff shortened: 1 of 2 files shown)


jonwright:
Looks great - I will merge it as is. There are a few things we do when doing this outside of the tk gui:
- applying the Lorentz correction
- a histogram of the number of spots (not weighted by intensity) gives the reflection multiplicities
- a histogram of log(intensity) is less biased by strong reflections
- pairing up spots first (g, -g) allows you to remove the effect of grain position in the sample

smerkel:
Maybe the histogram should allow to choose number of spots vs. peak intensity. 

As for lorentz correction, of log(intensity), I think that this is more at the post-processing stage and should probably not be built in the gui. This is like in most powder diffraction analysis where you get a raw intensity vs. 2theta.

What do you think?

jonwright:
The tkgui is a bit of a limitation nowadays, so I am not sure what to do about maintaining it? We have some help from DAU/ID06 at ESRF who are making some new workflow widgets (https://gitlab.esrf.fr/julia.garriga/imaged11_julia).

In the meantime, most of the scientists seem to be going over to jupyter notebooks. At ESRF there is a strong motivator that people can get onto our cluster via jupyter. Seems that jupyter is growing but it is not clear how to make a gui or get reasonable performance plotting lots of points.

For the powder patterns @younes-elhachi recently uploaded some code for doing powder patterns via a KDE in sandbox/weighted_kde/. Prior to that, we had a poster at an EPDIC a couple of years ago about producing these histograms. When we presented the poster Ken Beyerlein pointed us to his paper about the same idea : dx.doi.org/10.1017/S0885715614001171

I suspect there is a nice paper waiting to be written for using these kinds of spotty histogram patterns. Maybe building on Ida's work on particle statistics : https://doi.org/10.1107/S002188981102824X . I meant to try to write it but various things came up...

smerkel:
Not sure about jupyter notebooks for non experts. The tkgui is old and outdated, but accessible to starting students. Whether we use Tk / QT / anything doesn't matter to me but I like to have a simple bundle that people can download and run. 

Having a simple GUI is quite important if you want the technique to grow outside of the python geeks...

I do not know what would be the best solution.

PR #119, opened 2021-02-11
Numpy based histograms + weighting feature

smerkel:
I created a new function to compute twotheta histograms. Uses numpy and allow for weights. 

I would like, then, to add an option in transformer and the gui to choose whether to use number of peaks or peak intensity for calculating the histogram.

Would you be ok with this?

### DIFF
--- a/ImageD11src/transform.py
+++ b/ImageD11src/transform.py
@@ -352,6 +352,33 @@ def compute_tth_histo(tth, no_bins=100,
     # print "hpk",hpk[:10]
     return tthbin, histogram, hpk
 
+def compute_tth_histo_numpy(tth, no_bins=100, weight = False, weights = 0,
+                      **kwds):
+    """
+    Compute a histogram of tth values
+    Uses numpy's histogram rather that doing it by hand as above
+    New feature: weight by something (peak intensity for instance), send true for weight and weights values
+
+    Returns a normalised histogram (should make this a probability
+    *and*
+     For each datapoint, the corresponding histogram weight
+    2021-02-11 S. Merkel
+    """
+    maxtth = tth.max()
+    mintth = tth.min()
+    logging.debug("maxtth=%f , mintth=%f" % (maxtth, mintth))
+    if (weight):
+        histogram,binedges = np.histogram(tth, bins=no_bins, weights=weights, density=True)
+    else:
+        histogram,binedges = np.histogram(tth, bins=no_bins, density=True)
+    tthbin = np.mean(np.vstack([binedges[0:-1],binedges[1:]]), axis=0)
+    histogram = histogram/histogram.sum()
+    # histogram value for each peak
+    binsize = (maxtth - mintth) / (no_bins-1)
+    bins = np.floor((tth - mintth) / binsize).astype(np.int)
+    # print(len(bins), max(bins), min(bins), len(histogram))
+    hpk = np.take(histogram, bins) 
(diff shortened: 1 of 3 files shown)


jonwright:
I guess it makes sense to bundle something in which uses this code (or a testcase as least) to stop it getting broken in the future. I will try to add a few notes on the review as I didn't follow everything.

ImageD11src/transform.py
```diff
@@ -352,6 +352,33 @@ def compute_tth_histo(tth, no_bins=100,
     # print "hpk",hpk[:10]
     return tthbin, histogram, hpk
 
+def compute_tth_histo_numpy(tth, no_bins=100, weight = False, weights = 0,
```
jonwright, line 355: weights = None ?

ImageD11src/transform.py
```diff
@@ -352,6 +352,33 @@ def compute_tth_histo(tth, no_bins=100,
...
+    2021-02-11 S. Merkel
+    """
+    maxtth = tth.max()
+    mintth = tth.min()
+    logging.debug("maxtth=%f , mintth=%f" % (maxtth, mintth))
+    if (weight):
+        histogram,binedges = np.histogram(tth, bins=no_bins, weights=weights, density=True)
+    else:
+        histogram,binedges = np.histogram(tth, bins=no_bins, density=True)
+    tthbin = np.mean(np.vstack([binedges[0:-1],binedges[1:]]), axis=0)
```
jonwright, line 374: tthbin = 0.5 *(binedges[:-1] + binedges[1:])

ImageD11src/transform.py
```diff
@@ -352,6 +352,33 @@ def compute_tth_histo(tth, no_bins=100,
...
+    histogram = histogram/histogram.sum()
+    # histogram value for each peak
+    binsize = (maxtth - mintth) / (no_bins-1)
+    bins = np.floor((tth - mintth) / binsize).astype(np.int)
+    # print(len(bins), max(bins), min(bins), len(histogram))
+    hpk = np.take(histogram, bins)
```
jonwright, line 380: Add a comment here that len(bins) == total number of observed spots and len(histogram)  should be much less. Or use numpy.digitize ?

jonwright (commented):
It would make sense to clean up or remove the old code for this, which is clearly showing its age!

ImageD11src/transform.py
smerkel: Indeed!

ImageD11src/transform.py
smerkel: numpy.digitize would be better I think. I did not know the function existed.

ImageD11src/transform.py
smerkel: Ok.

smerkel:
I tested the new vs. the old code. The results are nearly identical except for the number of bins. The new function returns the exact requested number of bins. The old one returns the requested number of bins + 2.

If you prefer, I will push a new version of compute_tth_histo with the new code and remove my own function.

jonwright:
It seems better to clean it up to have just the good version. Thanks!!

smerkel:
I think it is ready now. I added a new parameter in transformer.py (weight_hist_intensities) to control whether to weight histograms by peak intensities. Small update to the tkgui so users can choose which type of histogram (number of peaks or peak intensities).

PR #142, opened 2022-02-19
Add pyproject.toml to avoid pre-installs with pip

AxelHenningsson:
An attempt to make `ImageD11`  installable with `pip` in a clean environment, without any prerequisites.

On a windows 10 using conda (as below) this installs with `pip` and passes `pytest` without any pre-installed packages such as `numpy` or the like. To replicate:

     conda create -n dev_ImageD11 python=3.8
     conda activate dev_ImageD11 
     git clone https://github.com/FABLE-3DXRD/ImageD11.git
     cd ImageD11
     pip install -e .
     conda install pytest
     pytest

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "six", "wheel", "numpy"]
+build-backend = "setuptools.build_meta"


jonwright:
I will merge as is but I have in mind that pytest should be coming in somehow automatically too?

AxelHenningsson:
Yeah I was surprised that the CI does not trigger tests on a pull requests. I have mostly used [GitHub actions](https://github.com/features/actions) to do CI stuff, so don't exactly know how the `.circleci` file works with tests in `ImageD11`.

PR #144, opened 2022-02-24
An attempt to fix pip install

jonwright:
It seems the pyproject.toml is somehow problematic with `pip install --user --editable`. Like this it seems to work, but needs testing to see if it still works for setting up an empty conda/venv.


jonwright:
@AxelHenningsson : can you let me know if this still works for your case?

PR #158, opened 2023-01-09
Update for numpy float and latest sinograms code

jonwright:


jonwright:
Updates, as well as the sad loss of np.float

PR #169, opened 2023-05-31
use the bg image, numpy deleted bools and numpy.float

haixing0a:
1) use the bg image in ImageD11/columnfile.py
2) numpy deleted bools in ImageD11/sinograms/lima_segmenter.py
3) change numpy.float to float in ImageD11/transformer.py


jonwright:
Thank you!

PR #171, opened 2023-06-28
Add diffraction origin peak updating to columnfile.py

AxelHenningsson:
This is a first suggestion for how the `columnfile.updateGeometry()` could be modified to make it possible for users that are interested in correcting their peak positions based on some belived diffraction origin positions rather than assuming that all peaks come from the origin as is currently assumed in  `columnfile.updateGeometry()`. Such corrections can be crucial for larger samples when recovering strains and lattice variations.

**Example:** *consider a scenario where we have a columnfile, `colf`, and a list of grains, `grains`, each featuring a centroid translation in x-lab coordinates `grains[i].dx` and a translation in  y-lab coordinates `grains[i].dy`. Additionally, each grain has a mask, `grains[i].mask`, indicating which peaks it has indexed in the `colf`. Then we may adopt something like the below to update the columnfile to better respect the fact that each peak is originating form a different part in the sample:*

```python
    colf.addcolumn( np.zeros((colf.nrows, )), 'gtx' )
    colf.addcolumn( np.zeros((colf.nrows, )), 'gty' )
    for g in grains:
        colf.gtx[g.mask] = g.dx
        colf.gty[g.mask] = g.dy
    s = np.sin(np.radians(colf.omega))
    c = np.cos(np.radians(colf.omega))
    colf.addcolumn( c*colf.gtx - s*colf.gty,  'dox' )
    colf.addcolumn( np.zeros((colf.nrows, )), 'doy' )
    colf.addcolumn( np.zeros((colf.nrows, )), 'doz' )
    colf.updateGeometry() # every peak is re-computed using an individual grain centroid...
 ```
*The idea is that once the columns `dox,doy,doz` have been set, specifying the diffraction origin (do) for each peak, the `columnfile.updateGeometry()` will respect this and compute the `tth,eta,gx,gy,gz` in accordance. In general one may set `dox, doy, doz` differently also within each `grains[i].mask` using information from the reconstructed grain shapes etc.*

### DIFF
--- a/ImageD11/columnfile.py
+++ b/ImageD11/columnfile.py
@@ -429,9 +429,16 @@ class columnfile(object):
         peaks_xyz = np.array((xl,yl,zl))
         assert "omega" in self.titles,"No omega column"
         om = self.omega *  float( pars.get("omegasign") )
-        tth, eta = transform.compute_tth_eta_from_xyz(
-            peaks_xyz, om,
-            **pars.parameters)
+        if "dox" in self.titles and "doy" in self.titles and "doz" in self.titles:
+            # respect diffraction origin positions, dox, doy, doz...
+            s1 = peaks_xyz - np.array((self.dox, self.doy, self.doz))
+            eta = np.degrees(np.arctan2(-s1[1, :], s1[2, :]))
+            s1_perp_x = np.sqrt(s1[1, :] * s1[1, :] + s1[2, :] * s1[2, :])
+            tth = np.degrees(np.arctan2(s1_perp_x, s1[0, :]))
+        else:
+            tth, eta = transform.compute_tth_eta_from_xyz(
+                peaks_xyz, om,
+                **pars.parameters)
         gx, gy, gz = transform.compute_g_vectors(
             tth, eta, om,
             wvln  = pars.get("wavelength"),

### CODE REVIEW

**Comment by AxelHenningsson** [comment]
[Time to abandon 2.7](https://github.com/actions/setup-python/issues/672) 😆 ?

**Comment by jonwright** [comment]
Hi Axel,

This patch looks a bit challenging for people to use without the recipe to compute dox, doy and doz (which is in the comment above rather than in the code). It seems to be a hidden special case, and I regret several others that have crept in over the years (e.g. space groups in transformer.py, negative cosine_tol, etc).

Would it also work for you to modify transform.py to just interpret the grain t_x, t_y and t_z parameters in the scanning setup rather than the 3D case? So `transform.compute_grain_origins` returns instead `(origin_x, 0, 0)` ? Then something like:
```
grain_peaks = colf.copyrows( g.mask )  # fix/debug this function if it doesn't work
grain_peaks.parameters.update( { 't_%s'%('xyz'[i]) : g.translation[i] for i in range(3) } )
grain_peaks.parameters.add( xfab.parameters.par('scanning3DXRD', 1) ) # or true?
grain_peak.updateGeometry()
```
Like this, I am vaguely optimistic that I can still use the forward projector in `compute_xyz_from_tth_eta`. It seems like the most useful way to get everything to work properly. The parameter name `scanning3DXRD` is not good yet. Maybe we could do better to give a beam size parameter in y/z and clip to that range. The beam size info is useful for a lot of other things anyway (e.g. beam_y, beam_z, same units as distance, pixels).

Another workaround option/idea for a quick hack: see which of the parameters can become `nrows` shaped arrays and still have the code run. For example, did you already try to just set:
```
c.parameters.parameters['distance'] = c.parameters.parameters['distance'] + c.dox 
```
It appears to work - dynamic typing is sometimes nice like that (but the C codes will fail). In that way, can you also add a per-peak sample-detector distance column and copy that into the parameters before calling update geometry? A dangerous hack could be to have any/all of the parameters check for the same-named columns to use instead of the scalar values. e.g.:

```
# ... inside updateGeometry
    colpars = { name : self.getcolumn[name] for name in self.parameters.parameters if name in self.titles } 
    self.parameters.parameter.update( colpars )
```

Fixing the consequences of this rule could be useful to allow merging of peaks when the detector is moving during a scan. It might be a big job...

For the 2.7 - it was waiting for someone to port FitAllB.

jonwright:
> [Time to abandon 2.7](https://github.com/actions/setup-python/issues/672) 😆 ?

fixed in #172

AxelHenningsson:
I do like the 
`c.parameters.parameters['distance'] = c.parameters.parameters['distance'] + c.dox` 
option you mention, it appears to work just fine. 👍 

I will close this issue then as the feature appears to not be needed. Rather a feature to get the "dox" to various degrees of accuracy is what should be supported imho.

Cheers
Axel

PR #180, opened 2023-10-31
Fix #177

### LINKED ISSUE #177: `ImageD11.cImageD11` and `multiprocessing.Pool` does not perform as expected
Setting `OMP_NUM_THREADS=1` and running cImageD11 functions in parallel with python native multiprocessing module does not perform as expected . The runtime per call increases (🤔?!) with the number of processes when it should stay exactly constant  as it would be for any single threaded pure python function. For instance `cImageD11.cImageD11.score_and_assign()` performs as below on some dummy data.

image

This is an issue when using ImageD11 for multi-grid indexing and various other more bizarre scanning-3dxrd applications. Perhaps there is a known trick to fix this, I know python native multiprocessing has been used a lot with ImageD11 so this should potentially be known? 🙃 

A self contained example below. (note that we are intentionally not measuring the Pool overhead of the forking and pickling-un-pickling) I have also tried to go `os.environ[OMP_NUM_THREADS]=1` with similar results. Similar things happens for other cImageD11 functions.

````python
import os
import numpy as np
import matplotlib.pyplot as plt
from multiprocessing import Pool
import ImageD11.cImageD11
from scipy.spatial.transform import Rotation
import ImageD11.grain
import time

def func(args):
    gve, grains, pin_omp_num_threads = args
    tol = 0.05
    drlv2 =  np.ones((gve.shape[1], ), 'd')
    label = -np.ones((gve.shape[1], ), 'i')
    if pin_omp_num_threads:
        ImageD11.cImageD11.cimaged11_omp_set_num_threads(1)
    t1 = time.perf_counter()
    for i, g in enumerate(grains):
            ImageD11.cImageD11.score_and_assign(g.ubi, gve.T, tol, drlv2, label, i)
    t2 = time.perf_counter()
    return (t2-t1)/len(grains)

if __name__ == "__main__":

    gve  = np.random.rand( 3, 100000 )
    grains = np.array([ImageD11.grain.grain(u) for u in Rotation.random(200).as_matrix()])

    plt.style.use('dark_background')
    plt.figure()
    for pin_omp_num_threads in [False, True]:
        time1core = func( (gve, grains, pin_omp_num_threads) )
        timings = [time1core]
        procs = [1]

        for nprocs in range(2, len(os.sched_getaffinity(0))):
            args = [(gve.copy(), gsub, pin_omp_num_threads) for gsub in np.array_split(grains.copy(), nprocs)]

            with Pool(processes=nprocs) as p:            
                runtimes = p.map( func, args )
            timings.append( np.mean(runtimes) )
            procs.append(nprocs)
            print(timings[-1])

        plt.plot(procs, timings, 'o--', label='pin OMP_NUM_THREADS : '+str(pin_omp_num_threads))
    plt.xlabel('Number of processes')
    plt.ylabel('Avg runtime [s/call] for ImageD11.cImageD11.score_and_assign()')
    plt.legend()
    plt.show()
````

jonwright:


jonwright:
I leave the 3.x test failing as it is waiting on a numba release for 3.12.

PR #181, opened 2023-10-31
pixel lut method fo Image distorsion correction 

jbjacob94:
added a new method to blobcorrector.correctorcalss to do image distorsion correction using a pixel LUT. Faster for large peakfiles


jonwright:
For some reason it is complaining about tabs? Can you run "black" on it and push again, please?
Thanks!

jbjacob94:
I edited the file with vim, and for some reasons it is making weird things with tabs. I updated the file so it should work now

PR #187, opened 2023-11-24
Add support for multiple detectors in sinograms

jadball:
- Add `guess_detector()` and `guess_motornames()` functions to `sinograms/datasets.py`
These functions read the `masterfile` and determine whether we are using `frelon3` or `eiger`, and what our `omegamotor` and `dtymotor` attributes should be (based on which motors are present in `instrument/positioners`).
They are called just before `import_scans()` and `import_motors_from_master()` respectively, to ensure we look for the correct data.
These functions are also called in the `load()` function - this is essential for `sinograms/lina_segmenter.py:setup()`
- Fixes `import_scans()` to handle broken softlinks (via a try/except cause to catch the `KeyError` thrown by `h5py`)

With these additions and fixes, the `NotesGuide` notebook from the gold example will complete successfully with my `frelon3` data.

### DIFF
--- a/.github/workflows/build_flake_pytest_ubuntu2004.yml
+++ b/.github/workflows/build_flake_pytest_ubuntu2004.yml
@@ -25,11 +25,12 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Ubuntu python 2.7
-      if: matrix.python == '2.7'
+      if: matrix.python-version == '2.7'
       run: |
           sudo apt-get update
-          sudo apt-get install -y --no-install-recommends python2 python3-virtualenv
+          sudo apt-get install -y --no-install-recommends python2 python2-dev python3-virtualenv
           virtualenv -p python2 ${HOME}/cp27
+          source ${HOME}/cp27/bin/activate
           ${HOME}/cp27/bin/python -m pip install -U pip
           ${HOME}/cp27/bin/python -m pip install -U setuptools wheel
           echo "${HOME}/cp27/bin" >> $GITHUB_PATH
@@ -37,12 +38,12 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install flake8 pytest
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        if [ -f requirements.txt ]; then python -m pip install -r requirements.txt; fi
         python -m pip install .
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
-        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics --exclude="*depreciated*"
+        python -m flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics --exclude="*depreciated*"
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         # flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
     - name: Test with pytest
--- a/ImageD11/cImageD11.py
+++ b/ImageD11/cImageD11.py
@@ -39,13 +39,15 @@ def check_multiprocessing():
     if hasattr(multiprocessing,"parent_process"):
         parent = multiprocessing.parent_process()
         # only for python 3.8 and up
-        if parent is not None:
+        if (parent is not None) and ('OMP_NUM_THREADS' not in os.environ):
             cimaged11_omp_set_num_threads( 1 ) 
         # people wanting Nprocs * Mthreads need to reset after import
-        # OMP_NUM_THREADS is not going to work for them
-        # how are we going to remember this in the future??
+        # or use OMP_NUM_THREADS as usual
     #
     # now check for the fork issue
+    if not hasattr(multiprocessing, 'get_start_method'):
+        # You are on python2.7. Give up.
+        return
     if ((multiprocessing.get_start_method(allow_none=False) == 'fork') and    # we have the problem
         (multiprocessing.get_start_method(allow_none=True) is None) and       # by accident
          ('forkserver' in multiprocessing.get_all_start_methods())):          # so fix it
--- a/ImageD11/sinograms/properties.py
+++ b/ImageD11/sinograms/properties.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 
 import os
 import logging
--- a/sandbox/sigma_3_matrices.py
+++ b/sandbox/sigma_3_matrices.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 
 import ImageD11.grain
 import ImageD11.grid_index_parallel
--- a/setup.py
+++ b/setup.py
@@ -157,6 +157,7 @@ useful = [   # stuff you probably want, and should be able to get easily
     'pyyaml',
     "pytest",       # for the CI
     'numba==0.46.0 ; python_version < "3" ',       # for some test cases
+    'llvmlite==0.30.0 ; python_version < "3" ',    # for some test cases
     'numba ; python_version > "3" ',               # for some test cases
     "bslz4_to_sparse",
     "fast_histogram",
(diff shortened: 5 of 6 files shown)


ImageD11/sinograms/dataset.py
```diff
@@ -20,6 +20,8 @@
 - indexing and reconstructing grains
 
 """
+
+POSSIBLE_DETECTOR_NAMES = ("frelon3", "eiger")
```
jonwright, line 24: We should just read the name from the dataset. There should be only one entry which has:

/scan/measurement/detector -> attrs[ "interpretation" ] == "image"

ImageD11/sinograms/assemble_label.py
```diff
@@ -13,8 +13,12 @@
 
 SCANMOTORS = ("diffrz", "diffrz_cen360", "diffrz_center", "fpico4", "fpico3", "diffty", "diffty_center",
               "rot", "rot_cen360", "rot_center", "fpico6", "dty", "dty_center")
-HEADERMOTORS = ("diffty", "diffrz", "samtx", "samty", "samtz", "diffry", "samrx", "samry",
-                "dty", "rot", "pz", "px", "py", "shtz", "shty", "shtx")
+#HEADERMOTORS = ("diffty", "diffrz", "samtx", "samty", "samtz", "diffry", "samrx", "samry",
+#                "dty", "rot", "pz", "px", "py", "shtz", "shty", "shtx")
+HEADERMOTORS_NSCOPE = ("dty", "rot", "pz", "px", "py", "shtz", "shty", "shtx")
+HEADERMOTORS_TDXRD = ("diffty", "diffrz", "samtx", "samty", "samtz", "diffry", "samrx", "samry")
+HEADERMOTORS = HEADERMOTORS_NSCOPE + HEADERMOTORS_TDXRD
```
jonwright, line 20: The intention of the single list was to take all potentially interesting names.

ImageD11/sinograms/dataset.py
```diff
@@ -85,7 +96,59 @@ def __repr__(self):
...
                 len(self.scans), self.scans[0], self.scans[-1] ))
         return  "\n".join(r)
+    
+    def guess_detector(self):
+        '''Guess which detector we are using from the masterfile'''
+        
+        # open the masterfile
+        hname = self.masterfile
+        detectors_seen = []
+        scan = "1.1"
```
jonwright, line 106: You need to import the scans first. Then find the detector for each scan.

ImageD11/sinograms/dataset.py
```diff
@@ -85,7 +96,59 @@ def __repr__(self):
...
 
+        with h5py.File( hname, 'r' ) as hin:
+            # go through the first scan, and see what detectors we can see
+            for measurement in list(hin[scan]['measurement']):
```
jonwright, line 110: for item in hin[scan]['measurement']:
    if "interpretation" in item.attrs and item.attrs[ "interpretation" ] == "image":
         ....

ImageD11/sinograms/dataset.py
```diff
@@ -85,7 +96,59 @@ def __repr__(self):
...
+                if measurement in POSSIBLE_DETECTOR_NAMES:
+                    detectors_seen.append(measurement)
+        
+        if len(detectors_seen) != 1:
```
jonwright, line 114: So the detector could be an argument to init to decide which one

ImageD11/sinograms/dataset.py
```diff
@@ -144,6 +207,8 @@ def import_all(self, scans=None, shape=None):
         
     def import_scans(self, scans=None, hname = None):
         """ Reads in the scans from the bliss master file """
+        # we need to work out what detector we have at this point
+        self.guess_detector()
```
jonwright, line 211: Dot we need one detector per scan? How does it work if someone does dct then far field in the same master file

ImageD11/sinograms/dataset.py
```diff
@@ -155,12 +220,16 @@ def import_scans(self, scans=None, hname = None):
...
-                    frames_per_scan.append( frames.shape[0] )
-                else:
+                try:
+                    frames = hin[scan]['measurement'][self.detector]
+                    if len(frames.shape)==3: # need 1D series of frames
+                        goodscans.append(scan)
+                        frames_per_scan.append( frames.shape[0] )
+                    else:
+                        print('Bad scan', scan)
+                except KeyError:
```
jonwright, line 230: I prefer this as: `if self.detector in scan:` rather than try/except if that is the intention.

ImageD11/sinograms/dataset.py
```diff
@@ -85,7 +96,59 @@ def __repr__(self):
...
+                if measurement in POSSIBLE_DETECTOR_NAMES:
+                    detectors_seen.append(measurement)
+        
+        if len(detectors_seen) != 1:
+            raise ValueError("More than one detector seen! Can't work out which one to process.")
+        else:
+            self.detector = detectors_seen[0]
+            
+        
+    def guess_motornames(self):
```
jonwright, line 120: To be seen - I think this should come from nexus style attributes or user input to scale in the future

ImageD11/sinograms/dataset.py
```diff
@@ -85,7 +96,59 @@ def __repr__(self):
...
+        using_nscope = False
+        using_tdxrd = False
+        for motor in motors_seen:
+            if motor in HEADERMOTORS_NSCOPE:
+                using_nscope = True
+            elif motor in HEADERMOTORS_TDXRD:
+                using_tdxrd = True
+        
+        if using_nscope and using_tdxrd:
+            raise ValueError("Found both nscope and tdxrd motors in positioners, not sure which one we were using!")
```
jonwright, line 143: This will not scale

ImageD11/sinograms/dataset.py
```diff
@@ -446,7 +517,13 @@ def load( self, h5name, h5group = '/' ):
 
 
 def load( h5name, h5group = '/' ):
-    return DataSet().load( h5name, h5group )
+    # return DataSet().load( h5name, h5group )
+    # make sure detector and rotation motors are known
+    ds_obj = DataSet().load( h5name, h5group )
+    ds_obj.guess_detector()
```
jonwright, line 523: Silently modifying the dataset when loading it from disk seems a bit dodgy. I prefer to have a method like `validate` or `repair` or `fix` that makes clear what is happening.

ImageD11/sinograms/dataset.py
```diff
@@ -47,9 +51,16 @@ def __init__(self,
                  dset = "dataset"):
         """ The things we need to know to process data """
 
+        
+        # it's better to determine the omegamotor, dtymotor and detector from the masterfile itself...
+        # however, that relies on realistic dataroot, analysisroot, etc values being passed to the init
+        # the default values for dataroot etc. will always give a broken masterfile path
```
jonwright, line 57: The default was downloading your data and working in the same folder as the dataset.

ImageD11/sinograms/dataset.py
```diff
@@ -47,9 +51,16 @@ def __init__(self,
                  dset = "dataset"):
         """ The things we need to know to process data """
 
+        
+        # it's better to determine the omegamotor, dtymotor and detector from the masterfile itself...
```
jonwright, line 55: Then have:
 ... ` omegamotor = None, dtymotor=None, detector=None `

`if thing is None: guess_thing()`

jonwright: If the info is supplied - the select the scan which have the motors.

Also handle the case of no dty motor.

Also handle the case of z scans vs y scans (etc, samry, samrx, diffry)

jonwright (commented):
Thanks!!!

So I have gone through reviewing here - it might be easier to do in person, but maybe it is useful to keep a log here to remember why things are being done. 

The overall idea for me would be : try to make it work in the future for some other beamline I have not seen yet. So less hardwiring of ID11 config.

jadball:
Good to discuss in person I think :)

jonwright:
Quick thought this morning - it would be good to keep your guess_detector logic in a `import_from_blissfile` function that is similar to import_from_sparse.

PR #189, opened 2023-11-30
Enable manually specifying detector and motor names

jadball:
Updated version of pull request #187

You can now specify `detector`, `omegamotor` and `dtymotor` in the `__init__` method of the `DataSet` class - this enables support for other detectors such as the frelon.


jonwright:
Great - thanks!

PR #192, opened 2023-12-07
Add new solver for strain and stress 

jbjacob94:
Create a new solver to compute strain and stress from ImageD11 ubis

### DIFF
--- a/ImageD11/eps_sig_solver_NEW.py
+++ b/ImageD11/eps_sig_solver_NEW.py
@@ -1,0 +1,669 @@
+import numpy as np
+import ImageD11.finite_strain, ImageD11.unitcell, ImageD11.parameters
+from ImageD11.grain import read_grain_file
+
+
+# Compute strain and stress from a list of ImageD11 UBI matrices 
+# Copyright (C) 2023 Jean-Baptiste Jacob
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


jonwright:
Thanks!!

I think we can merge this as it is and fix the python2.7 CI later (I guess it is just reformatting the f-strings).

Pinging @jadball and @AxelHenningsson as they have similar codes and might have good ideas for creating some testcases to verify the numbers and keep things working into the future

AxelHenningsson:
Hi! It would be great with some Context? Is there any new maths here?

Cheers 
Axel

jonwright:
The conversion from strain to stress ought to match the old code in FitAllB. I think the invariants and eigen decompositions are new as python code here, but the maths should be known. A missing link (for me) is to have a verifiable check whether the numbers come out right for some low symmetry cases (e.g. quartz for trigonal vs hexagonal). There are some factors of 2 and sqrt(2) that are not obvious.

jbjacob94:
Hi!


nothing new with the math indeed, I just tried to remove any dependency with old python 2 code and I added a few functions to get the invariants, eigen decomposition and deviatoric component of strain / stress.


> I try to stick to the IEEE conventions for stiffness matrices. Is that what is used here? I think a link in the constructor the the convention would be good.


I did not know there was a convention, but good to know. Do you have a link to send? I used voigt notation for the stiffness matrix, which seems to be commonly used, at least in the mineralogy community.


best

JB

________________________________

PR #193, opened 2023-12-08
Bugfix assemble_label.py

jonwright:
set rather than list randomly permuted the order of scans. Oups.


jonwright:
This still needs to be tested with an fscan2d example (e.g. gold) for the scan::num business.

PR #195, opened 2023-12-11
Fixes for python2.7

jonwright:
@jbjacob94 This fixes the CI so that the code can run on python2.7, and that will allow us (eventually) to test it against fitallb.


jadball:
I'll make a new PR to fix the f-string in stress.py

PR #206, opened 2024-01-24
Resurrect tests for columnfile.py, correct np.int

jadball:
In the Github Actions CI, the command "python -m pytest" is run to test ImageD11.
By default, this only runs test files with the naming convention `test_*.py`
Renamed `testcolumnfile.py` to `test_columnfile.py` to fix this.
Also, replaced `np.int` references `int` to fix numpy depreciation errors.


jadball:
@jonwright happy to merge?

PR #208, opened 2024-01-26
Parallelised iradon

jadball:
I was able to parallelise the iradon function.
This was achieved by distributing the interpolation calls across multiple processes.
I've also added a check to use the (supposedly faster) `np.interp` method instead of the scipy one if the desired interpolation is linear.
With 7 workers (seems to be the fastest for the shepp-logan phantom) and 3600 projections, this new parallelised version is around 6 times faster than before.

### DIFF
--- a/ImageD11/sinograms/roi_iradon.py
+++ b/ImageD11/sinograms/roi_iradon.py
@@ -93,16 +93,16 @@ def _get_fourier_filter(size, filter_name):
 
 
 # skimage.transform.iradon
-def iradon(radon_image, theta, 
+def iradon(radon_image,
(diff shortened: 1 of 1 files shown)


jonwright:
Thanks! I will merge it like this. This code you have added is parallel over projections and it is doing a reduction into the final image (so a lot of summing on a single thread, or threaded by numpy?). 
The original idea was to chop the output image into tiles (or rows) and then treat each one in a thread. So then there is no locking.
I think this can still improve this approach if each thread sums several projections and returns only one frame to be summed (e.g. do some of the reduction summation in parallel).

PR #209, opened 2024-02-01
dataset.py file structure enforcement and MLEM mask

jadball:
Apologies for many lines of changes, my code editor auto-linted things.
The changes to `dataset.py` are outlined:

1. Added `update_paths()` method to `DataSet` class
This function determines `self.analysispath` and a set of paths to directories and files such as `self.pksfile` and `self.grainsfile`
This function is called during `self.__init__()` but also at the end of `self.load()` because the paths depend on `self.sample` and `self.dataset` which are not defined when the main `load()` function is called (because an empty `DataSet()` object is created and afterwards loaded from file).
2. Added a default h5 file path to the `self.save()` and `self.load()` methods.
If no h5 path is manually provided, `h5name` is set to `self.dsfile_default` which is calculated in `self.update_paths()`
At the end of `self.save()` and `self.load()`, we know the file saved successfully, so we update `self.dsfile` equal to `h5name` (so we can keep track of where we saved to)

Only one actual change to `roi_iradon.py` :

1. Added mask capability to MLEM function
This mask is not given to the initial `sensitivity_image` iradon call because we invert the result which means dividing by zero.
The mask is instead just applied right at the end before returning `mlem_rec`
This avoids weird corner streak artifacts observed when running MLEM by 15+ iterations.

### DIFF
--- a/ImageD11/sinograms/dataset.py
+++ b/ImageD11/sinograms/dataset.py
@@ -92,13 +92,15 @@ class DataSet:
         # root of analysis for this dataset for this sample:
         self.analysispath = os.path.join(self.analysisroot, self.sample, self.dsname)
 
-        self.dsfile_default = os.path.join(self.analysispath, 'dataset.h5')
+        self.dsfile_default = os.path.join(self.analysispath, self.dsname + '_dataset.h5')
         # at the moment, set self.dsfile to be the default
         # if save or load is ever called, this will be replaced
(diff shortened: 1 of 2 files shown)


jonwright:
Looks good to me, Thanks! 
For the formatting - it would be great to figure out how to add "black" as a commit hook so that this gets sorted out magically.

PR #210, opened 2024-02-02
Update release continuous integration

t20100:
This PR:
- Updates `cibuildwheel` action to fix an issue on Windows (see https://github.com/pypa/cibuildwheel/releases/tag/v2.16.5)
- Updates `upload|download-artifacts` actions: This now allows to download the produced wheel/tarball from the action page (see https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
           platforms: all
 
       - name: Build wheels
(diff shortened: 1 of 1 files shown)


jonwright:
Thanks! I need to fix a bug in guicommand and then ask you to update the ESRF installation again ...

PR #225, opened 2024-02-21
Scanning and box-beam 3DXRD notebooks, and a few other changes

jadball:
### **CHANGES OUTSIDE NBGUI**
- Corrected typo in sinograms/assemble_label.py Line 212
- Added `self.col3dfile` to sinograms/dataset.py
- Removed `method = multiprocessing.set_start_method("forkserver")` from sinograms/point_by_point.py
- Pad `projection_shifts` in MLEM recon

### **ADDED BOX BEAM NOTEBOOKS**
- Similar to scanning notebooks but for box-beam 3DXRD
- Currently only working for Frelon data
- Uses Frelon segmenter from ma4750 (in sandbox) for peaksearching
- Global parameter refinement notebook
- Uses makemap.py to refine grain positions
- Some extra notebooks for importing multiple slices in a z series and looking at peak shapes (still under development)

### **SCANNING NOTEBOOK CHANGES**

General changes:
- Moved quite a few common functions into utils.py. These need properly incorporating into ImageD11 in the future
- All notebooks updated to obey DataSet filestructures for peaks and grains files

0_S3DXRD_segment:
- Added widget written by Pierre-Olivier for adjusting segmenter parameters
- Added slurm bash "listener" that waits for slurm job to complete before next cell completes
- Moves individual sparse files to their own subfolder after assembling

1_S3DXRD_index:
- More sensible indexer that prioritises low-multiplicity rings
- Grains now saved in HDF5 format
- Added another example notebook for indexing a minor phase

2_S3DXRD_sinograms:
- Added whole-sample sinogram reconstruction to use as a mask for individual grain reconstructions.
- Added MLEM refinement stage (uses SLURM) to grain reconstructions using run_mlem_recon.py
- Added another example notebook for mapping a minor phase (uses grain recons for centre-of-mass)

3_S3DXRD_strain_maps_pbp:
- Performs a point-by-point refinement of the grain-averaged UBI using 2D peak selection
- Uses grain labels array from sinograms for grain shapes
- Should give "relative" Type-III strains

4_S3DXRD_plot_both_phases:
- Small example notebook for plotting both phases in the same map
- Some initial work on orientation relationship identification using Orix (not finished yet)


jadball:
There's quite a bit to do still, but I'm conscious of the size of the changes already, so I thought I would PR what I had for now and continue development. My next primary task will be to sort out `utils.py` (at the very least, combine them between box-beam and scanning).

jonwright:
I will merge as is, but there are a few things to be checked

PR #238, opened 2024-03-04
Local point-by-point indexing notebook

jadball:
- Cleaned up and incorporated `ImageD11/nbGui/run_pbp.ipynb` as a scanning notebook: `ImageD11/nbGui/S3DXRD/1_2_3_S3DXRD_pbp_indexing_local.ipynb`
- Added notebook to view the results:  `ImageD11/nbGui/S3DXRD/4_S3DXRD_view_PBP.ipynb`
- Many tasks still to do but it works, and it's fast!


jonwright:
Why is this deleting the run_pbp notebook?

jonwright:
I think I am missing a tool to view a notebook with diffs from git / github. The json is not readable at all. There is stuff in run_pbp that was correct ($HOME) that is a going to be a problem in the new notebook outside of the cluster (/home/esrf/$USER).

jadball:
@jonwright This just pulls the run_pbp notebook into the Scanning subdirectory with the rest of the notebooks and cleans up the paths to the dataset etc.

jadball:
Problem with $HOME should be fixed in #240

PR #240, opened 2024-03-04
Improve home directory determination in PBP notebooks

jadball:
Now uses $HOME rather than $USER :)


jadball:
Fixes problem pointed out in #238

PR #242, opened 2024-03-06
Optionally scale sinogram rows by ring current

jadball:
Added ring current scaling to sinograms, removes reconstruction artifacts caused by inconsistent peak intensities between rotations


jonwright:
The (fpico6, fpico4) monitor should be imported to the sparse files during assemble_label. We can add an option somewhere to use this in properties.py or add it to the peaks table (then normalise during pk2d or pk2dmerge). For ring current I thought it is in the scan 1.2 groups and you need to interpolate it somehow to get the right data shape.

For this pull request, I think ds.ring_currents has shape NYxNOMEGA but ssino has shape NHKL ??

jonwright:
If you want to correct the whole rows with an average ring current (or monitor) per scan, then it needs to be a bit more clear how it works, I was assuming this would correct frame by frame.

jadball:
In `map_grain_from_peaks` in the notebook, `ssino` always has `NY` rows:
```python
NY = len(ds.ybincens)  # number of y translations
[...]
sino = np.zeros( ( npks, NY ), 'f' )
hits = np.zeros( ( npks, NY ), 'f' )
angs = np.zeros( ( npks, NY ), 'f' )
adr = destRow * NY + iy 
# Just accumulate 
sig = flt.sum_intensity
ImageD11.cImageD11.put_incr64( sino, adr, sig )
ImageD11.cImageD11.put_incr64( hits, adr, np.ones(len(de),dtype='f'))
ImageD11.cImageD11.put_incr64( angs, adr, flt.omega)

sinoangles = angs.sum( axis = 1) / hits.sum( axis = 1 )
# Normalise:
sino = (sino.T/sino.max( axis=1 )).T
# Sort (cosmetic):
order = np.lexsort((np.arange(npks), sinoangles))
sinoangles = sinoangles[order]
ssino = sino[order].T
```

At the moment, I am correcting each Y translation by the ring current. Looking at a few different datasets over the last few years, we always have `instrument/machine/current` available in the `.1` scans. Depending on the macro used to collect the data, we sometimes don't have `.2` scans.

Going forward, I agree it would be better to correct each frame - I will make an issue for this.

jadball:
Hold off on merging for now, I'll make a correction function for the sinogram with documentation so it's clearer what this does

jonwright:
Feel free to merge if you prefer as it is. My feeling is that doing a monitor correction when building the sinogram seems a bit busy.  You can do this normalisation either before (to the peaks) or after (to the sinogram) without needing it to be part of the peaks-to-sinogram builder function.

jadball:
I think what we have now is a good hold-over until we are properly scaling the peaks, I'll add a new function to correct the sinograms after they are created rather than during creation

PR #261, opened 2024-03-18
Moving many functions from nbgui into ImageD11

jadball:
Apologies for the monster commit! Key highlights below:

- New GrainSinogram class to handle building sinograms from a `Grain` and `DataSet` object.
- HDF5 load/save methods for `Grain` class
- Many useful helper functions in `DataSet` class (e.g getting peaks tables for you)
- Some utility functions inside `iradon`
- Lima segmenter now uses mask and background files from the dataset object if none are manually provided
- Utility function for `blobcorrector.py` to correct a columnfile with a spline

All the diffs should be readable aside from `grain.py` which I accidentally linted (sorry!) - look at the bottom of the file for new stuff added.

### DIFF
--- a/ImageD11/blobcorrector.py
+++ b/ImageD11/blobcorrector.py
@@ -318,7 +318,19 @@ class eiger_spatial(object):
         s = self.dx.shape
         i, j = numpy.mgrid[ 0:s[0], 0:s[1] ]
         return self.dy + j, self.dx + i
-    
+
+
+def correct_cf_with_spline(cf, spline_file):
+    """Creates a correctorclass from the spline file
+       Corrects the columnfile with the spline file
+       Returns the corrected columnfile"""
+
+    corrector = correctorclass(spline_file)
+    corrector.correct_px_lut(cf)
+
+    return cf
+
+
 #
 #"""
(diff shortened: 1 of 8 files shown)


jadball:
This seems to be failing due to a missing hdf5plugin on py 2.7 - @jonwright any ideas? I can push a new commit in this PR to fix if you know how

jadball:
I think the problem is that `dataset.py` now imports `properties.py` which needs `hdf5plugin` for its IO functions. We are only just now seeing this issue on Python 2.7 because we have some tests in `test_datasets.py` that are now indirectly targeting code with HDF5 I/O

jadball:
This may be solved by adding hdf5plugin as an explicit dependency to ImageD11?

jonwright:
We should have hdf5plugin in setup.py already:
     'hdf5plugin==1.4.1 ; python_version < "3" and sys_platform == "win32" ',
     'hdf5plugin ; python_version >= "3" ',

jonwright:
But I guess there is a bug there as it has nothing for python2 on non-windows...

jonwright:
I will merge and attempt to fix the CI later, the problem is not related...

PR #263, opened 2024-03-18
fix CI

jonwright:


jonwright:
Looks more complicated. There is no shared memory on 2.7. Probably a re-factor is needed to stop using shared memory anyway.

PR #274, opened 2024-03-28
reduce atol/rtol on tests

jonwright:
Sorry - Black format messed up the diff. Main changes:
- the color tests have a smaller tolerance (orientation was wrong for blue)
- added a check on the code path of 1 UB vs many UBs


jonwright:
@jadball OK?

jadball:
@jonwright looks good, thanks for checking this!

PR #275, opened 2024-03-28
Remove implied units (microns vs mm for dty vs diffty)

jonwright:
@jadball OK? Again black has trashed the formatting. The first diff might be cleaner if there is a way to see it.


jadball:
Happy, thanks for reviewing! Merging now

PR #276, opened 2024-04-11
Geometry overhaul (+tests), update all notebooks

jadball:
- Refactor of `ImageD11.sinograms.geometry`
    - Now tested in `test/test_geometry.py`
    - Updated `ImageD11.sinograms.point_by_point` accordingly
        - This has been validated against lab geometry and sinograms method for the Al foil + wire sample
    - Demonstration notebook available in `sandbox/scanning_geometry.ipynb`
- Added `get_cf_3d_from_disk()` method to `ImageD11.sinograms.dataset`
- Updated shift mechanics in `run_iradon()` and `run_mlem()` inside `ImageD11.sinograms.roi_iradon`
    - Replaced `y0` parameter with `shift` (because we don't shift by `y0` anyway)
    - Removed errant negative sign in the shift call
    - Updated `ImageD11.sinograms.sinogram` to match this
- Fixed some HDF5 IO bugs in `ImageD11.sinograms.sinogram`


jonwright:
I will merge, but there are some things to look at. In point_by_point it was intentional to precompute the cos/sin omega before doing the dtymask selection for each point...

PR #284, opened 2024-05-28
Release workflow: Increase verbosity of cibuildwheel

t20100:
Following some recent troubleshooting by @kif with this, this PR proposes to increase the verbosity of the build.

It also removes the `--pre` installation flag which makes the tests of the generated source tarball fail because it run the tests with numpy v2.

BTW, if you want to support numpy v2, the wheels should be generated with numpy v2.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -87,7 +87,7 @@ jobs:
 
       - name: Install and test sdist
         run: |
-          pip install --pre dist/imaged11*.tar.gz
+          pip install dist/imaged11*.tar.gz
           pytest test
 
       - uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


jonwright:
I am confused about the tar ball being lower case? The last one uploaded on pypi was mixed case matching the project. Has something changed for that?

t20100:
I'll have to check, I though you changed something.

t20100:
> I am confused about the tar ball being lower case? The last one uploaded on pypi was mixed case matching the project. Has something changed for that?

This changed in setuptools v69.3 to standardize package naming.

https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode : 

> In distribution names, any run of -_. characters (HYPHEN-MINUS, LOW LINE and FULL STOP) should be replaced with _ (LOW LINE), and uppercase characters should be replaced with corresponding lowercase ones.

jonwright:
OK - I will merge and hope we have the right setuptools to match.

t20100:
The latest version of `setuptools` is used, so it should be an issue.

Issue #297, opened 2024-07-11
Default values with update_path() in dataset.py


CharlesRmn:
Fix to load the right default values. Otherwise the col3dfile expects sample/sample_dataset/sample_dataset_peaks_3d.h5.

```
def update_paths(self):
    # paths for processed data
    # root of analysis for this dataset for this sample:
    self.analysispath = os.path.join(self.analysisroot, self.sample, self.dsname)

    self.dsfile_default = os.path.join(
        self.analysispath, self.dsname + "_dataset.h5"
    )
    # at the moment, set self.dsfile to be the default
    # if save or load is ever called, this will be replaced
    self.dsfile = self.dsfile_default
    # They should be saved / loaded with the dataset.
    for name, extn in [
        ("pksfile", "_peaks_table.h5"),
        ("col4dfile", "_peaks_4d.h5"),
        ("col3dfile", "_peaks_3d.h5"),
        ("col3dfile", "_peaks_2d.h5"),
        ("grainsfile", "_grains.h5"),
        ("sparsefile", "_sparse.h5"),
        # subset peaks selected for indexing (pbp)
        ("icolfile", "_icolf.h5"),
        # point by point raw output
        ("pbpfile", "_pbp.txt"),
    ]:
        # If the user has got a different name (via loading or edit), we keep that
        if getattr(self, name, f'./sample/sample_dataset/sample_dataset{extn}') == f'./sample/sample_dataset/sample_dataset{extn}':
            # Otherwise, these are the defaults.
            setattr(self, name, os.path.join(self.analysispath, self.dsname + extn))
```

jonwright:
Is the problem fixed by the code in #299?

Otherwise, maybe the loading code is the problem.

I am not so happy about keeping this "sample_dataset" default in the code.

jonwright:
Fixed via #299

jadball:
Right now I see the following:

https://github.com/FABLE-3DXRD/ImageD11/blob/1d6d7173851642090358fc2f2a241f8e32db8ebd/ImageD11/sinograms/dataset.py#L148-L150

Shouldn't this be:

```python
            ("col4dfile", "_peaks_4d.h5"),
            ("col3dfile", "_peaks_3d.h5"),
            ("col2dfile", "_peaks_2d.h5"),
```

jonwright:
yes - looks like a a typo?

jonwright:
Hopefully fixed by the incoming pull request.

PR #299, opened 2024-07-12
Load via the init method because of the update paths bug

jonwright:
This might work around #297


CharlesRmn:
It does the job indeed !

PR #306, opened 2024-07-23
Compacted the 0_S3DXRD notebook

jonwright:
Still needs the bootstrap to go into some other project?


jonwright:
Fix the bootstrapping issue another day. It will not be needed if we update ImageD11 in the site-wide install.

PR #310, opened 2024-07-24
Many many scanning notebook changes

jadball:
Individual commits describe roughly what was done - should all be updated with new Python path stuff, better plots, much better MLEM reconstructions using ASTRA, + some other various quality of life improvements


jadball:
Ready to merge when you are :)

PR #311, opened 2024-07-24
silicon calibration for nscope

jonwright:


jonwright:
So I think the fail is a side effect of `dataset.parfile = None`. I will merge this anyway as it is too late today to continue.

PR #337, opened 2024-10-15
do not square the sum_intensity

jonwright:


jonwright:
Could start processing the gold wire data on a laptop from an ssd

PR #338, opened 2024-10-17
Update release workflow to use macos-13

t20100:
This PR updates the release workflow to:
- use `macos-13` (11 is no longer available and 12 is deprecated)
- fix the installation of dependencies for the tests to pass

Here is a manual run of it: https://github.com/FABLE-3DXRD/ImageD11/actions/runs/11384434752


jadball:
Thanks Thomas! LGTM

Issue #325, opened 2024-09-19
Multi-tensor map class


jadball:
We need a new class (or extend `PBPMap` from https://github.com/FABLE-3DXRD/ImageD11/blob/8c895dd4a54863a87b5692852afcb71e45d7d890/ImageD11/sinograms/point_by_point.py#L113) to handle point-by-point refinement.

The class needs to support the following conditions:

- [ ] Outer dimensions should be in reconstruction space (`ri`, `rj` like we use everywhere else)
- [ ] Support mixed-type, ragged-length arrays:
    - [ ] At each map pixel, we have variable-length `ubis`, `eps` and `npks` depending on how many grains we found at that point
- [ ] Not necessarily mutable (could make a new array for refinement results)
- [ ] IO support **essential** - we have to be able to save to disk
- [ ] Some `selbest` methods - for generating single-valued maps from multi-valued maps. A number of different methods could be used here.
- [ ] Generation from a `TensorMap` - results of tomographic refinement

From what I can see, `AwkwardArray` seems like a good way to start:
https://awkward-array.org/doc/main/getting-started/index.html

@jonwright what do you think?

jadball:
I worry that having both a "SoA" carrying the raw data and an awkward array for row selections makes things very complicated. I think that the "SoA" approach could suffer a lot from slowdown when we are asking to get all the data at a specific grid point. Doing something like `grain_map[50, 30]` should always be pretty quick, and doing this as a columnfile would mean getting all the rows that match `i=50 & j=30` which I fear would be much slower.  Limiting the scope of our problem to a many-valued regular grid would simplify things a lot?

jonwright:
The AoS vs SoA question is whether to store the columnfile table data by rows or by columns (e.g. doesn't matter on paper, depends on how computers work, etc). So long as the table rows are sorted by position in space then it can be fast/easy to index into them for both layouts.

With column storage, for the series of different arrays like UBI, UB, eps, sigma, etc, we should be able to share the "position in space indexing" pointers across all columns. Probably this can be done via any one of awkward/ragged/pydata_sparse_gcxs. Or we just have one array for each data item like  `UBI[ngrains, 3, 3]` and for each point in space we have the `start`, `len` indices to grab `UBI[ start[i,j] : start[i,j] + len[i:j] ]` (probably fails for `len == 0`). So in class PBPmap we would need to:

- ensure the data are grouped by position in space (no need to sort, but all grains from one position should be together)
[...]

Issue #278, opened 2024-04-23
GrainMap class


jadball:
We need a class to store information about a spatial map of a grain/multiple grains.
This class should do at least the following:

- [x] I/O to HDF5 group
- [x] Calculation of the following arrays:
    - [x] Grain ID - by combining recons from multiple different grains together
    - [x] Per-grain UBI - from grain objects
    - [x] Per-voxel UBI - by local refinement
    - [x] IPF X, Y, Z
    - [ ] Misorientation
    - [ ] Grain ID (segmentation from point-by-point data?)
    - [x] Phase ID?
 - [x] Creation methods:
    - [x] From a list of `GrainSinogram` objects
    - [x] Form the UBI array directly from point-by-point indexing

jadball:
Mostly completed by #322 - per-voxel UBI by local refinement is ongoing, and I'm not sure we should be segmenting point-by-point data (leave as a task for MTEX?)

jonwright:
About MTEX - the main problem seems to be output of multi-channel versus single channel results. Does MTEX allow several orientations per point together with a weight for each one?

jadball:
No, as far as I am aware MTEX is single-valued. I have implemented a MTEX CTF export as a `TensorMap` classmethod.

jadball:
Solved by #339

jadball:
Closing

PR #339, opened 2024-10-17
Scanning 3DXRD Strain refinement

jadball:
# Summary:
We can now do point-by-point strain refinement for S3DXRD data, by implementing a method provided by @AxelHenningsson  . Thanks for your help!
This works for grain maps coming from both the point-by-point indexing process and the tomographic mapping process.

## PBPRefine:
This is a class to manage the point-by-point refinement process, similar to the `PBP` class.
Like the PBP class, we choose some peaks for `self.icolf` which we will refine with.
The class also has an input multi-valued map (`self.pbpmap`), a refinement mask (`self.mask`) and a single-valued 'choice' of the multi-valued map (required for computing x origins for each grain) (`self.singlemap`). For now this is computed from `PBPMap.choose_best`.
Features:
- Save and load from HDF5
- `self.get_origins()`: Compute origin-of-diffraction for each peak in `self.icolf`
- `self.run_refine()`: Initiate point-by-point refinement, either locally or on a cluster machine through SLURM
Solves #278

## PBPMap:
This is now a subclass of `ImageD11.columnfile.columnfile`. This solves the confusions with multi-valued data structures, ragged arrays, etc etc. in a very simple way that we were already using for the PBP refinement results
- Added plotting methods for nuniq histogram and best results (from notebooks)
- Added strain support (for strain refinement methods)
- Solves #325

## PBP:
- Now computes corrected origin of diffraction for g-vectors to account for x-position in the sample

## Other bits and bobs:

- Updated Zenodo dataset with multiphase pars (updated `fetch_data.py` and its test accordingly)
- Renamed scanning notebooks to more clearly define the choice of processing routes (and how they diverge then merge together later)
- Added some more documentation to the scanning notebooks in the form of markdown text blocks
- Added some new utility functions to `ImageD11/sinograms/geometry.py`
- Added some more stress and strain functions to `ImageD11/sinograms/tensor_map.py`

## TODO:

- Automated tests that run explicitly against @AxelHenningsson 's implementation (during development this was being done semi-manually)
- Test the functions I had to grab from ImageD11 to Numba-fy against their references
- Test stress stuff in `TensorMap`


jadball:
Python 2 tests are currently failing because `roi_iradon` is trying to import `scipy.fft` and failing

jadball:
@jonwright are you happy to merge as-is and we exclude `point_by_point` tests on Python 2.7? I think this is the first time our CI touches the `roi_iradon` code which is why we haven't seen this before.

jadball:
Got it, merging now! Managed to skip the test.

PR #341, opened 2024-10-18
Add forward_model.py to facilitate peak filtering, sinogram cleaning and cf set difference, together with a few auxillary tools for interacting with tomo, admet and orientation conversion

haixing0a:
Add forward_model.py to facilitate peak filtering, sinogram cleaning and cf set difference, together with a few auxillary tools for interacting with tomo, admet and orientation conversion


jonwright:
Thanks @haixing0a  - I will try to help to fix the python2.7 syntax problems before merging it. For now, I want to keep ImageD11 compatible with FitAllB etc that were not ported to python3 yet.

PR #342, opened 2024-10-23
Use xl, yl, zl for origin of diffraction calculation in PBP indexer

jadball:
@jonwright can you please double-check the maths behind this? I believe I am correctly modifying the distance but it'd be good to double-check: the new function `get_local_gv` does this for the indexer.

### DIFF
--- a/ImageD11/sinograms/point_by_point.py
+++ b/ImageD11/sinograms/point_by_point.py
@@ -28,8 +28,8 @@ import numba
 
 # ignore Numba dot product performance warnings?
 import warnings
-warnings.simplefilter('ignore', category=numba.core.errors.NumbaPerformanceWarning)
 
+warnings.simplefilter('ignore', category=numba.core.errors.NumbaPerformanceWarning)
 
 import pprint
 
@@ -79,7 +79,6 @@ def hkluniq(ubi, gx, gy, gz, eta, m, tol, hmax):
         if hcount.flat[i] > 0:
             ucount += 1
     return tcount, ucount
-
 
 # stuff we need to compute g-vectors
 # from ImageD11.transform
@@ -1047,7 +1046,7 @@ class PBPRefine:
 
         if guess_speed:
             guess_npks = min(guess_npks, self.icolf.nrows)
-            print('Running test function twice for the first',guess_npks, 'peaks to guess speed...')
+            print('Running test function twice for the first', guess_npks, 'peaks to guess speed...')
             print('This is because we get speed advantages with peaks on consecutive frames')
 
             idx = np.arange(guess_npks)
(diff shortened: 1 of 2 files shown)


ImageD11/sinograms/point_by_point.py
```diff
@@ -1651,42 +1696,30 @@ def idxpoint(
...
+    # then score with peaks[m]
+    # so we need to compute g-vectors for peaks[m]
+    # then mask those for indexing via [mr]
+
+    # compute g-vectors for peaks[m]
+    gv, gx, gy, gz = get_local_gv(i, j, ystep, omega[m], sinomega[m], cosomega[m], xl[m], yl[m], zl[m])
+    eta_local = eta[m]
+
+    # mask g-vectors to [m & mr]
+    gv_local_mask = gv[local_mask[m], :]
```
jonwright, line 1716: The comment suggests this should be: `gv[ m & mr , : ]`  or `gv[ local_mask, : ]`. Not sure why there is an [m] on the mask?

ImageD11/sinograms/point_by_point.py
```diff
@@ -25,6 +25,12 @@
 
 import h5py
 import numba
+
+# ignore Numba dot product performance warnings?
+import warnings
+
+warnings.simplefilter('ignore', category=numba.core.errors.NumbaPerformanceWarning)
```
jonwright, line 32: Rather than suppressing the message, we should eventually fix the actual problem here

ImageD11/sinograms/point_by_point.py
```diff
@@ -1602,6 +1615,36 @@ def refine_map(refine_points, all_pbpmap_ubis, ri_col, rj_col, sx_grid, sy_grid,
...
+    # compute local offset
+    xl_local = xl - x_offset
+    # stack xl, yl, zl
+    xyz_local = np.column_stack((xl_local, yl, zl))
+    # hold computed g-vectors
+    gv = np.empty((len(cosomega), 3))
+    # compute g-vectors from xyz
+    ImageD11.cImageD11.compute_gv(xyz_local,
+                                  omega,
+                                  parglobal.get('omegasign'),
```
jonwright, line 1635: parglobal is not defined inside this function

ImageD11/sinograms/tensor_map.py
```diff
@@ -1,6 +1,10 @@
 import os
 
 import numba
+# ignore Numba dot product performance warnings?
+import warnings
+warnings.simplefilter('ignore', category=numba.core.errors.NumbaPerformanceWarning)
```
jonwright, line 6: I would like to understand why it gives the warning

jonwright (commented):
I didn't understand the mask arg. Otherwise, for the sign check I think the only way is just to plot the silicon data (dstar error vs omega) to check if it works.

ImageD11/sinograms/point_by_point.py
jadball: As `parglobal` is globally defined, I don't think it needs to be defined inside this function? It seems to run on my machine, anyway...

ImageD11/sinograms/point_by_point.py
jadball: My thought process went like this:

We want to calculate the g-vectors for all the peaks at this pixel (mask `m`).
However, we want to index only the g-vectors that are also in `isel` (mask `mr`).

We compute `gv` (and `gx, gy, gz`) from `peaks[m]` (the `dtyi, omega` mask) so `len(gv) = m.sum() != len(sinomega)`.

However `mr` has the wrong length to directly mask `gv`.
If we combine `local_mask = m & mr` first, then we can mask `local_mask` by `m` to get the `mr` mask for only the `m`-masked values, so the masks have the right lengths.
Therefore `gv[local_mask[m], :]` should be correct.

jadball:
> Otherwise, for the sign check I think the only way is just to plot the silicon data (dstar error vs omega) to check if it works.

This might be tricky - we're now directly computing the g-vectors from `xl, yl, zl`, so I'm not sure how we'd access the modified `dstar` values - unless we can compute `dstar`/`tth` back from the new g-vectors?

jonwright:
Dstar is the length of gv. I guess the function can be tested outside of the call here.

ImageD11/sinograms/point_by_point.py
jonwright: If you want the ring mask at this pixel I would expect to read that as  `mr[ local_mask ]` ?

ImageD11/sinograms/tensor_map.py
jonwright: If we suppress the warning we should create an issue to come back to it sometime in the future. Like this I am afraid we lose all of the feedback on numba compilation for anything importing our code...

ImageD11/sinograms/point_by_point.py
jadball: The way I see it is that we are creating a (ring and pixel) mask for all the peaks using `local_mask = m & mr`, then selecting a (pixel) subset of that with `local_mask[m]` which then has the same length to index `gv` (which was also created from `peaks[m]`).

jadball: Perhaps the problem is just unclear variable naming?

`m`: Mask on all peaks to select those that come from the pixel
`mr`: Mask on all peaks to select those on the ring (`isel`)
`local_mask`: Mask on all peaks to select those on the ring AND those from the pixel

jonwright: It seems like it can also be `gv[ mr[m] ]` with a comment like `# selection: mr = active on rings, [m] at this point in space`.

The thing you have is:
`gv[ ( m & mr )[m] ]`

While I see this is mathematically equivalent, having the m in there twice meant I could not figure out what you were doing.

Feel free to merge and I can always change it later.

jadball:
Just tested this against the previous version and it should be equivalent - should be good to merge

PR #345, opened 2024-10-29
Fix strain to stress conversion in TensorMap class, better TensorMap documentation

### LINKED ISSUE #194: Stress/Strain elastic constants testing and validation
Moving the discussion in merged pull #192 into a new issue. The problem is which numerical values to input for the numbers in the calculations of stress from strain, and whether the numbers can be checked.

- Some unit tests could be added.

- we do not have the old FitAllB strain convention in ImageD11 yet. This could be added for checking against fitallb.

Then for trigonal quartz as a low symmetry example. We had an indexing ambiguity, the (100) direction is not distinguished from the (110) direction until you look at the peak intensities. For high-temperature beta quartz, the structure is hexagonal. For stresses in alpha quartz, it seems to come down to the sign of C14, which goes to zero when the structure goes from trigonal to hexagonal (Ohno, I., Harada, K. & Yoshitomi, C. Temperature variation of elastic constants of quartz across the α - β transition. Phys Chem Minerals 33, 1–9 (2006). https://doi.org/10.1007/s00269-005-0008-3). For crystals that are twinned, this might not matter.

There is a description of the problem of conventions here:
https://www.comsol.com/blogs/piezoelectric-materials-understanding-standards/

In the case of quartz, the C14 parameter appears to change sign when going from the IRE 1949 convention to the one from IEEE 1978. Some intensity ratios are discussed here:
https://royalsocietypublishing.org/doi/pdf/10.1098/rspa.1926.0025
... and their relation to macroscopic crystal axes is here:
http://www.minsocam.org/ammin/AM30/AM30_326.pdf
... and here, BSTJ 22: 3. October 1943: Use of X-Rays for Determining the Orientation of Quartz Crystals. Bond, W.L.; Armstrong, E.J.:
https://archive.org/details/bstj22-3-293

Somehow we need to summarise all that literature into something that matches a set of elastic constants to a crystal structure that gives the diffracted intensities. Maybe taking the numbers from the materials project is unambiguous, as they are providing both together:

https://next-gen.materialsproject.org/materials/mp-7000?crystal_system=Trigonal&formula=SiO2

jadball:
We now compute the Biot stress tensor in the crystal reference frame (ImageD11 convention), given a Biot strain tensor in the crystal reference frame (ImageD11 convention, the default obtained when calling `ImageD11.grain.eps_crystal_matrix(dzero_cell)`) and a 6x6 stiffness tensor (Voigt-compatible, IEEE 1976 lattice convention, what [materialsproject.org](https://next-gen.materialsproject.org/) gives you).

Thanks to @AxelHenningsson and @jonwright for discussions on this.
Partially fixes #194 - the dream is to move the `Numba` kernels from `tensor_map.py` into `stress.py` eventually


ImageD11/sinograms/tensor_map.py
```diff
@@ -212,32 +313,136 @@ def ubi_and_unitcell_to_eps_crystal(ubi, dzero_cell, res):
...
+        res[...] = np.nan
+    elif np.isnan(B0[0, 0]):
+        res[...] = np.nan
+    elif not phase_mask:
+        res[...] = np.nan
+    else:
+        # rotate the strain tensor from ImageD11 coordinate system to IEEE 1976 coordinate system
+        # determine rotation matrix E
+        C_c = np.linalg.inv(B0).T
+        E = np.column_stack((C_c[:, 0], np.cross(C_c[:, 2], C_c[:, 0]), C_c[:, 2]))
```
jonwright, line 409: Looks extremely dodgy. Not a rotation.

jonwright:
Sure - go for it.

PR #355, opened 2024-11-19
fix geometry to match module and cleanup and fit a CeO2

jonwright:


jonwright:
Ignoring the python2.7 failure for now - it is not related to this pull request

PR #356, opened 2024-11-21
Consolidated powder and SX calibration notebooks

jadball:
- Combined mine and @jonwright powder calibration notebooks together
- Modernised a bit the single-crystal calibration notebook
- Moved them both to a 'calibration' sub-folder
- Removed some older not-useful calibration notebooks
- Updated `transformer` class a bit so we can use it without saving a `.flt` to disk


jadball:
Don't merge yet - TODO: split index_unknown notebook into two (recon method and friedel method)

Issue #352, opened 2024-11-15
TensorMap slow import


jonwright:
```
%%time
from ImageD11.sinograms.tensor_map import TensorMap
CPU times: user 8.83 s, sys: 242 ms, total: 9.07 s
Wall time: 9.47 s
```

Adding cache=True somewhere might help (https://stackoverflow.com/questions/44131691/how-to-clear-cache-or-force-recompilation-in-numba).

jadball:
Resolved by @CiolJegou #375

PR #375, opened 2025-01-13
 TensorMap slow import #352 

CiolJegou:
Adding (cache = True) in the numba occurences 
On first import :
```
CPU times: user 9.63 s, sys: 390 ms, total: 10 s
Wall time: 12 s
```
On second import (after restarting the kernel) :
```
CPU times: user 716 ms, sys: 298 ms, total: 1.01 s
Wall time: 960 ms
```


jonwright:
Thanks Loïc! Can you just remove the auto-generated _cImageD11 files? These are not needed in git and they are different for each python/numpy version so I prefer not to commit them.

jadball:
Thanks! Also please apply `cache=True` to Numba functions inside `ImageD11/sinograms/point_by_point.py` if possible :)

Issue #366, opened 2024-12-13
pbp indexer - y0 not added in grid and units for diffty not matching distance


jonwright:
Around line 2000 in point_by_point:
"""
        nlo = np.floor((self.ybincens.min()) / self.ystep).astype(int)
        nhi = np.ceil((self.ybincens.max()) / self.ystep).astype(int)

        rng = range(nlo, nhi + 1, gridstep)
"""
It needs to subtract y0 if there was one.

Also noted: the distance correction in get_local_gv assumes dty and distance have the same units.  This is harder to fix automagically unless everything comes with units.

jonwright:
Also the indexing loglevel does not work when changed. Something about the multiprocessing pool and global state perhaps. No idea how to fix that.

jadball:
For the first issue: here's the commit to blame: https://github.com/FABLE-3DXRD/ImageD11/commit/32dfd946ab5ee80f1a285606b41f65824e8fbc32

It seems I was correcting a `y0` bug for a dataset with a big y0 by modifying the code in two places - I remember things definitely got better after the bug fix, but I suspect the second fix wasn't needed. To me this raises deeper questions about how the `geometry.dty_to_dtyi` function works/should work. TODO after Christmas for sure!

jadball:
I think the underlying problem is that we assume that a perfectly aligned diffractometer would have a y0 value of 0?

jadball:
The simplest fix for this (without possibly introducing more bugs by modifying `geometry.py`) would be to have an option to shift `self.dty` in the `DataSet` class when the motor values are read/imported, to account for where the rotation axis should be in the lab frame assuming perfect alignment (~14 for the TDXRD station, 0 for the NSCOPE station). @jonwright what do you think? The alternative is I think some rather drastic additions to `geometry.py` to introduce a new static reference frame centered on the diffractometer.

jonwright:
I didn't understand why the 14 is different to y0 yet? I will try to go through the code soon.

Just adding some known offset to the silicon testcase should give something small to look at numbers and debug...

jonwright:
Is it just:

Sinogram scan step space != Recon step space

?

jadball:
> Is it just:
> 
> Sinogram scan step space != Recon step space
> 
> ?

@jonwright you're right.
Continuous `(dty, omega)` space should have its own discretisation.
By convention we use `(ystep, ostep)`.
[...]

PR #379, opened 2025-01-16
Geometry fixes

jadball:
Addresses #366

Highlights:

# dty units of microns now enforced in `dataset.py`
Microns are overwhelmingly our choice for `.par` files and I think other solutions introduce significant potentials for confusion.
I solve this by checking the `BLISS` axis units for `dtymotor` and converting from mm to microns if needed:

```python
def import_motors_from_master(self):
    dty_is_mm = False
            (...)
            # read from h5:
            dty = hin[scan]["instrument/positioners"][self.dtymotor]
            # get the unit for dty:
            dty_unit = hin[scan]["instrument/positioners"][self.dtymotor].attrs['units']
            if dty_unit == 'mm':
                dty_is_mm = True
            (...)
    # ensure dty is in microns - this is for compatability with geometry
    if dty_is_mm:
        self.dty = self.dty * 1000
        logging.info("Converted dty from mm to um")
    logging.info("imported omega/dty")
```

# Introduced a new geometric parameter `ybeam`
This is the nominal y-value of the beam in the lab frame. Usually 0 for NSCOPE, 14 mm for TDXRD.
Old approach:

```python
def sample_to_lab_sincos(sx, sy, y0, dty, sinomega, cosomega):
    # first, rotate sx and sy by omega about its origin (rotation axis)
    sxr = sx * cosomega - sy * sinomega
    syr = sx * sinomega + sy * cosomega

    # then translate about the rotated frame
    lx = sxr
    ly = syr + dty - y0

    return lx, ly
```

New approach:
```python
def sample_to_lab_sincos(sx, sy, y0, ybeam, dty, sinomega, cosomega):
    # first, rotate sx and sy by omega about its origin (rotation axis)
    sxr = sx * cosomega - sy * sinomega
    syr = sx * sinomega + sy * cosomega

    # then translate about the rotated frame
    lx = sxr
    ly = syr + dty + (ybeam - y0)

    return lx, ly
```

I have updated the rest of `geometry.py` to match, including tests and Jupyter notebooks.
This affects less than you might think, because `ybeam` often cancels out and we rarely use the lab frame in computation.

# Updated `geometry.py` function names and docstrings for clarity
# Decoupled `dtyi` (discretisation of `dty`) from "step space" `(si, sj)`
New `dty` to `dtyi` conversion function now requires `ymin` but is conceptually simpler and **does not induce a sign flip any more**:

```python
def dty_to_dtyi(dty, ystep, ymin):
    """
    We take continuous dty space and discretise it
    We do this by counting the number of ysteps away from ymin
    ymin should probably be ds.ymin
    This should be equivalent to determining the closest bin in ds.ybincens
    But this is neater as it accounts for values outside the ds.ybincens range

    dtyi = (dty - ymin) / ystep
    """
    dtyi = np.round((dty - ymin) / ystep).astype(int)
    return dtyi
```
I've updated the `point_by_point.py` code accordingly to include `ymin` and `ybeam` where needed.
`ybeam` is basically only needed when considering shifts in tomographic reconstructions.

# Introduced new standard way to generate grids in `(si, sj)` to reconstruct on
This is `geometry.step_grid_from_ybincens` which I discuss here: https://github.com/FABLE-3DXRD/ImageD11/issues/366#issuecomment-2593561198


jadball:
Caveat here: we don't store the units for `dty` in sparse files...

jonwright:
A couple of quick comments:

- I don't understand the difference between ybeam and y0. Why not just say y0=14 for TDXRD? It seems to add a redundant parameter.

- the units thing should be for a new version and not in this release. Old datasets have no dty units and I used mm for distance and pixel size a few times in the past. Forcing microns now seems to break the fitting I did with Ryan.

jadball:
> * I don't understand the difference between ybeam and y0. Why not just say y0=14 for TDXRD? It seems to add a redundant parameter.

You're right - I now understand this better. If we define the lab frame via the beam (e.g `ly = 0` when a grain is in the beam rather than `ly = ybeam`) then it all resolves without issue.

The final sticking point is the `shift` calculation for the tomographic reconstruction, which I think I've nailed now. New commits incoming!

ImageD11/sinograms/dataset.py
```diff
@@ -418,6 +429,10 @@ def import_motors_from_master(self):
                 print(
                     "replace bad scan omega", b, self.scans[b], "with", j, self.scans[j]
                 )
+        # ensure dty is in microns - this is for compatability with geometry
+        if dty_is_mm:
+            self.dty = self.dty * 1000
```
jonwright, line 434: To modify the dty numbers is a breaking change. It should not go into the next release.

jadball:
New highlights:

# Updated `geometry.py` function names and docstrings for clarity
# Decoupled `dtyi` (discretisation of `dty`) from "step space" `(si, sj)`
New `dty` to `dtyi` conversion function now requires `ymin` but is conceptually simpler and **does not induce a sign flip any more**:

```python
def dty_to_dtyi(dty, ystep, ymin):
    """
    We take continuous dty space and discretise it
    We do this by counting the number of ysteps away from ymin
    ymin should probably be ds.ymin
    This should be equivalent to determining the closest bin in ds.ybincens
    But this is neater as it accounts for values outside the ds.ybincens range

    dtyi = (dty - ymin) / ystep
    """
    dtyi = np.round((dty - ymin) / ystep).astype(int)
    return dtyi
```
I've updated the `point_by_point.py` code accordingly

# Introduced new standard way to generate grids in `(si, sj)` to reconstruct on
This is `geometry.step_grid_from_ybincens` which I discuss here: https://github.com/FABLE-3DXRD/ImageD11/issues/366#issuecomment-2593561198

# Accounting for `shift` and `pad` properly in tomographic reconstructions:
```python
def sino_shift_and_pad(y0, ny, ymin, ystep):
    """
    Determine the difference in sinogram pixels between
    the central dtyi value of the sinogram and the dtyi value of the rotation axis (from y0)
    To do this, we convert y0 into dtyi without rounding
    Also determine the minimum pad to get the whole sample in the frame
    Which should be double the shift
    """
    # get the middle row of the sinogram in pixel space
    ymid_px = ny/2
    # get the value of y0 in pixel space
    # this is the same as dty_to_dtyi without rounding
    y0_px = (y0 - ymin)/ystep
    shift = ymid_px - y0_px
    pad = np.ceil(np.abs(shift) * 2).astype(int) + 1
    return shift, pad
```

jadball:
@jonwright ready when you are if you're happy to merge :)

PR #387, opened 2025-01-23
Move SegmenterGui utils into a separate file

loichuder:
The import of functions from `segment_gui` raises a `ModuleNotFoundError: ipywidgets` with a regular install of ImageD11 (i.e. no `more` or `full` extra requirements)

These utils actually do not depend on `ipywidgets`: we should be able to expose them with a regular installation.

To solve this, I moved the utils to a separate file. I kept the compatiblity with the existing notebooks by reimporting the utils functions in `segmenter_gui` (so that, for example, `segmenter_gui.guess_ESRF_paths()` still works)

### DIFF
--- a/ImageD11/nbGui/segmenter_gui.py
+++ b/ImageD11/nbGui/segmenter_gui.py
@@ -1,4 +1,3 @@
-import os
 import numpy as np
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


jonwright:
Looks good to me? The 2.7 CI should be fixed by the other incoming patch, so I will merge this.

PR #393, opened 2025-02-04
add forward_projector, grainmaps and updates with forward_model etc to forward compute

haixing0a:
The main changes are:
1) add forward_projector.py, take the experimentally reconstructed tensor map (from pbp_tensor_refined or tensor map in xx_grains.h5), define the beam profile and forward compute peaks and output them as cf_2d, cf_3d and cf_4d, which have the same formats as ImageD11 column file. The goal is to check the peaks difference and bridge with DCT, and hopefully helps to develop iterative forward and back projections;
2) add grainmaps.py, main features are i) merge regions when the voxels have similar orientations, 'similar' is defined by calculating the misorientation to the mean orientation of the region, therefore a min_misorientation is defined by the user to identify grain IDs (called 'labels'); ii) indexing_iterative for match peaks with already-indexed grains, then use the remaining peaks to index new grains and the obtained new grains will be merged with already-indexed grains to remove duplication
3) updates with forward_model etc, to add logging, improve computational speed etc and add testing notebooks

### DIFF
--- a/ImageD11/forward_model/__init__.py
+++ b/ImageD11/forward_model/__init__.py
@@ -1,12 +1,13 @@
 """
-forward model for 3DXRD and s3DXRD
+forward model for box-beam 3DXRD and s3DXRD using point-focused beam
 The aim is to exploit forward model for the following purposes:
 1) facilitate analysis for (s)3DXRD by filtering and cleaning peaks for indexed grains, calculating completeness
 2) bridging the combined analysis between DCT and (s)3DXRD for the same sample but measured in different geometries/instruments
 3) forward model based reconstruction for both near-field and far-field geometries
 4) parameters conversion among ImageD11 par, DCT parameters, poni, which all can be converted to a dictionary and compatible with fwd-DCT parameters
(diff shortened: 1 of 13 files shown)


jadball:
some of the stuff (like the disorientation code) we may already have in ImageD11 in the `sym_u` library - I think it could also be possible to combine the `grain_map` and `TensorMap` classes by just adding new the new grain merging methods to `TensorMap`.

For now, though, I suggest merging as-is, then we can work on integrating it more with existing code in the future

haixing0a:
Yes, I fully agree. In the future, it would be nice to see TensorMap has the options: i) to merge regions for rendering effective 'labels' (grainIDs); ii) remove noisy grains and update the grain IDs; iii) compute the completeness so one knows what grain is more trustable and what is less. These options are right now available in grainmaps.py, but better to be integrated with TensorMap. 

Also this "indexing_iterative" should be in the future go together with some of the grain indexing functions. I think we can integrate better after we do more testings on these codes. In the future, I would be happy to see grainmaps.py gone after it is integrated well with other functionalities.

PR #394, opened 2025-02-06
small fixes with peak merging for forward_projector

haixing0a:
small fixes with 4D peak merging for forward_projector and other minor updates


jonwright:
Thanks! Do you have write access to merge it yourself in here? If so then please go ahead, if not then let's figure out how to set that up...

PR #395, opened 2025-02-07
frelon_peaksearch process function split into two functions

abmajith:
Hello @loichuder, @jadball  

I split the frelon_peak search process function into two, first for running the segment on frames in parallel,
and then merging the peaks found on each frames, so on.

### DIFF
--- a/ImageD11/frelon_peaksearch.py
+++ b/ImageD11/frelon_peaksearch.py
@@ -360,6 +360,95 @@ PKSAVE = [
 PKCOL = [getattr(ImageD11.cImageD11, p) for p in PKSAVE]
 
 
+def collect_all_frames_peaks(
+    dataset,
+    worker_args,
+    num_cpus=None,
+    **process_map_kwargs,
+):
+    """
+    Collects peaks for all the frames in the first scan in the dataset 
+    using parallel processing.
+    """
+    num_threads = num_cpus or max(1, ImageD11.cImageD11.cores_available() - 1)
+    master_file = dataset.masterfile
+    scan_name = dataset.scans[0]
+    frames_dataset = f"{scan_name}/measurement/{dataset.detector}"
+    omega_angles = dataset.omega[0, :]
+
+    num_frames = omega_angles.shape[0]
+    args = [
+        (master_file, frames_dataset, i, omega_angles[i], worker_args) 
+        for i in range(num_frames)
+    ]
+    all_frames_peaks_list = process_map(
+        pps, 
+        args, 
+        chunksize=1, 
+        max_workers=num_threads, 
+        **process_map_kwargs
+    )
+    return all_frames_peaks_list
+
+def merge_all_frames_peaks(all_frames_peaks_list):
+    """
+    Merges collected peak properties into structured dictionaries
+    """
+    peaks_2d_dict = {title: [] for title in PKSAVE}
+    num_peaks = []
+
+    for peak_index, properties in all_frames_peaks_list:
+        for title, column_index in zip(PKSAVE, PKCOL):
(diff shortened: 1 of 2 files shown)


ImageD11/frelon_peaksearch.py
```diff
@@ -360,6 +360,95 @@ def guess_bg(ds, scan_number=0, start=0, step=9, n=None):
 PKCOL = [getattr(ImageD11.cImageD11, p) for p in PKSAVE]
 
 
+def collect_all_frames_peaks(
+    dataset,
```
loichuder, line 364: Perhaps a bit early but as @jadball suggested, we should not need the `Dataset` class.

We can give the `masterfile`, frames_dataset` and `omega_angles` as arguments of the function so that we can reuse it without further changes when we remove the use of the Dataset class.

ImageD11/frelon_peaksearch.py
```diff
@@ -360,6 +360,95 @@ def guess_bg(ds, scan_number=0, start=0, step=9, n=None):
...
+        num_peaks.append(properties.shape[0])
+    
+    # Convert lists to numpy arrays
+    for key in peaks_2d_dict:
+        peaks_2d_dict[key] = np.array(peaks_2d_dict[key])
+    
+    return peaks_2d_dict, num_peaks
+
+
+def process_dataset_for_peaks_columnfile(
```
loichuder, line 412: This is essentialy a reimplementation of the `process` function below right ? 

Should we use the new functions directly in the `process` function instead of creating a new `process` function ? What do you think @jadball ?

abmajith: yes, that is true,

ImageD11/frelon_peaksearch.py
abmajith: correct, this will make future compatible!

jonwright (commented):
Maybe I missed some discussion prior to this? The 'importScan.py' code looks like an extraction/refactoring of dataset.py but it is not imported or used in frelon_peaksearch.py. Is this intended to be part of this PR?

Your type hints for CI seem to require https://peps.python.org/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code

abmajith:
> Maybe I missed some discussion prior to this? The 'importScan.py' code looks like an extraction/refactoring of dataset.py but it is not imported or used in frelon_peaksearch.py. Is this intended to be part of this PR?
> 
> Your type hints for CI seem to require https://peps.python.org/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code

abmajith:
> Maybe I missed some discussion prior to this? The 'importScan.py' code looks like an extraction/refactoring of dataset.py but it is not imported or used in frelon_peaksearch.py. Is this intended to be part of this PR?

I wrote this Import_scan to do the segmentation with minimal *dataset* class method but subset of *dataset* class method, to make the columnfile treatable from its own class, instead of doing it in *dataset* class

jonwright:
No problem - if I were you, I would be tempted to run the segmentation starting just from the bliss "master.h5 + scan" and see if you can drop everything else from our dataset class. It was intended for assembling 2D sinograms from many 1D scans...

loichuder:
>  The 'importScan.py' code looks like an extraction/refactoring of dataset.py but it is not imported or used in frelon_peaksearch.py. Is this intended to be part of this PR?

Yes. I don't think it was intended for this PR.

@abmajith As discussed this week, you don't have to close the PR: keep 633012dca51ab441576f3a4b787a4446083a2499 only, which the commit that we discussed, in this branch. 

You could then keep the other commits in another branch in the meantime.

PR #396, opened 2025-02-07
frelon_peaksearch process function split into two functions 

abmajith:
@jonwright, @jadball @loichuder


ImageD11/frelon_peaksearch.py
```diff
@@ -360,6 +360,100 @@ def guess_bg(ds, scan_number=0, start=0, step=9, n=None):
...
 
 
+def collect_all_frames_peaks(
+    master_file,
+    scans,
+    omega_angles,
+    detector,
+    worker_args,
+    num_cpus=None,
+    **process_map_kwargs,
```
jonwright, line 370: Lint for 2.7 seems to dislike the trailing comma, which is only allowed since 3.6: https://test-black.readthedocs.io/en/style-guide/style_guide/trailing_commas.html

ImageD11/frelon_peaksearch.py
```diff
@@ -360,6 +360,100 @@ def guess_bg(ds, scan_number=0, start=0, step=9, n=None):
...
+):
+    """
+    Collects peaks for all the frames in the first scan in the dataset 
+    using parallel processing.
+    """
+    num_threads = num_cpus or max(1, ImageD11.cImageD11.cores_available() - 1)
+    scan_name = scans[0]
+    frames_dataset = f"{scan_name}/measurement/{detector}"
```
loichuder, line 378: Maybe I missed something but `scans` and `detector` seem to be used only to compute `frames_dataset`.

Meaning you could pass `frames_dataset` instead of `scans` and `detector`, reducing the number of parameters.

Similarly, only the `omega_angles` that are relevant to the scan could be passed to the function (i.e. pass `omega_angles[0]` instead of the full `omega_angles`.

This would make the function available for any scan, not only the first one if we so desire.

ImageD11/frelon_peaksearch.py
```diff
@@ -360,6 +360,94 @@ def guess_bg(ds, scan_number=0, start=0, step=9, n=None):
...
+    """
+    Process a dataset by collecting all 2d peaks (for frames),
+    merge themn,
+    and performing spaital correction (detector plane correction)
+    Returns columnfile_2d, and columnfile_3d
+    """
+    # Step 1: collect all peaks
+    all_frames_peaks_list = collect_all_frames_peaks(
+        dataset.masterfile,
+        f"{dataset.scans[0]}/measurement/{dataset.detector}",
```
jonwright, line 424: This one is:
"%s/measurement/%s"%(dataset.scans[0], dataset.detector)

... although I suspect this will not work for a dataset with 2 scans (either side of the load frame) as someone will want to [process(scan) for scan in dataset.scans]. Historically we did two segmentations and merged the peaks later. @jadball may know more about whether that is possible?

abmajith: Yes, as of now the code only works for only one scan

jonwright: Can you add the scan number as an optional argument (scan_number=0) for both the process function and this one? Otherwise we have to modify dataset.scans to be able to use this code to peak search a dataset that has two or more scans.

loichuder:
I'll reiterate my comment from #395: should we not replace `process` by `process_dataset_for_peaks_columnfile` since it does the same thing ? Or did I miss something ?

Otherwise, the code looks fine to me.

abmajith:
> I'll reiterate my comment from #395: should we not replace `process` by `process_dataset_for_peaks_columnfile` since it does the same thing ? Or did I miss something ?
> 
> Otherwise, the code looks fine to me.

I'm just waiting for @jadball  to confirm o.w I could have removed it

jonwright (changes requested):
There is a bunch of other cleanup to be done in here, to discuss with @jadball. It is not needed for this MR, but if you are putting this into production for EWOKS it would be better to have something we will not want to edit again soon.
- 3d merge is not merging most of the 2D columns if we are keeping all of PKSAVE. All of the columns can be merged, e.g: https://github.com/FABLE-3DXRD/ImageD11/blob/24810fb9871cb9480dbda7bfaf0b03643502b47c/src/blobs.c#L133
- the dataset cache looks buggy under multiprocessing fork (hdf5 limitation)
- the plots method should either plot something or be removed

abmajith:
@jonwright Thanks for your input! I realized the change you pointed out after merging, but to avoid a last-minute rush over the weekend, I decided to put it on hold.

As you mentioned, I had considered this approach. I think the best way forward is to keep the current branch with minor modifications as a separate fork. This way, we can work incrementally on the Ewoks task, making it easier for both me and @loichuder to develop incrementally and maintain it.

Regarding your suggestion, I'll check with @jadball and also with Pierre-Olivier. I recall that his tomography master file for segmentation looks similar to what we have in 3DXRD. 
Based on the availability of James and Pierre-Olivier will work refine the *dataset* class, simplify things where possible, and provide a clear specification. Also about the 3D merging in blob.c file as you mentioned. We will handle this work in separate branch and test it extensively.

ImageD11/frelon_peaksearch.py
```diff
@@ -423,7 +423,7 @@ def process_dataset_for_peaks_columnfile(
     all_frames_peaks_list = collect_all_frames_peaks(
         dataset.masterfile,
         "%s/measurement/%s"%(dataset.scans[scan_number], dataset.detector),
-        dataset.omega[0,:],
+        dataset.omega[scan_number,:],
```
jonwright, line 426: Please delete the  `,:` and have `dataset.omega[scan_number]` so that we can have a list of omega in dataset (in the future)

ImageD11/frelon_peaksearch.py
```diff
@@ -466,7 +466,7 @@ def process(ds, worker_args, ncpu=None, scan_number=0, **process_map_kwargs):
     hname = ds.masterfile
     scan_name = ds.scans[scan_number]
     frames_dset = scan_name + "/measurement/" + ds.detector
-    omega = ds.omega[0, :]
+    omega = ds.omega[scan_number, :]
```
jonwright, line 469: Same comment about ",:"

(Sorry to be pedantic - on line 438 you have this)

jonwright (changes requested):

jadball:
> > I'll reiterate my comment from #395: should we not replace `process` by `process_dataset_for_peaks_columnfile` since it does the same thing ? Or did I miss something ?
> > Otherwise, the code looks fine to me.
> 
> I'm just waiting for @jadball to confirm o.w I could have removed it

I think that's fine. I will suggest a few name changes of functions for clarity

jadball (changes requested):

ImageD11/frelon_peaksearch.py
```diff
@@ -360,7 +360,96 @@ def guess_bg(ds, scan_number=0, start=0, step=9, n=None):
...
+    all_frames_peaks_list = process_map(
+        pps, 
+        args, 
+        chunksize=1, 
+        max_workers=num_threads, 
+        **process_map_kwargs
+    )
+    return all_frames_peaks_list
+
+def merge_all_frames_peaks(all_frames_peaks_list):
```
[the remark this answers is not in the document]
abmajith: I think combine or club will be action, like club_all_frames_peaks or combine_all_frames_peaks,

ImageD11/frelon_peaksearch.py
```diff
@@ -360,7 +360,96 @@ def guess_bg(ds, scan_number=0, start=0, step=9, n=None):
...
+    peaks_2d_dict.pop("o_raw")
+    columnfile_2d = dataset.get_colfile_from_peaks_dict(peaks_2d_dict)
+
+    peak_3d_dict["spot3d_id"] = np.arange(len(peak_3d_dict["s_raw"]))
+    columnfile_3d = dataset.get_colfile_from_peaks_dict(peak_3d_dict)
+
+    return columnfile_2d, columnfile_3d
+
+
+def process(ds, worker_args, ncpu=None, scan_number=0, **process_map_kwargs):
```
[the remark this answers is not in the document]
abmajith: I will rename the process_dataset_for_peaks_columnfile with process, and remove the earlier process function, to make the notebook continue to work without change

ImageD11/frelon_peaksearch.py
```diff
@@ -360,7 +360,96 @@ def guess_bg(ds, scan_number=0, start=0, step=9, n=None):
...
+        num_peaks.append(properties.shape[0])
+    
+    # Convert lists to numpy arrays
+    for key in peaks_2d_dict:
+        peaks_2d_dict[key] = np.array(peaks_2d_dict[key])
+    
+    return peaks_2d_dict, num_peaks
+
+
+def process_dataset_for_peaks_columnfile(
```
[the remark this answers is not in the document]
abmajith: okay, yes

ImageD11/frelon_peaksearch.py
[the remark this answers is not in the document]
jonwright: I am now imagining a caveman smashing the computer with a club :laughing: 

Whichever names are chosen, I am happy for this to get merged.

Thanks!

ImageD11/frelon_peaksearch.py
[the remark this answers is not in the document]
abmajith: Hey James (@jadball),  to rename this process function into segment_dataset, I need to replace process function name into segment_dataset name in your "0_3DXRD_segment_frelon.ipynb" notebook  to make it work

I wrote it and will request you for the merge

PR #398, opened 2025-02-12
created a logic for filter_peaks_by_phase  in peakselect.py file

abmajith:
@loichuder 

In order to do the phase based peaks filtering and intensity based peak filtering, I created a filter_peaks_by_intensity in peakselect.py file in imageD11,

### DIFF
--- a/ImageD11/peakselect.py
+++ b/ImageD11/peakselect.py
@@ -1,7 +1,8 @@
 # coding: utf-8
 
 from __future__ import print_function, division
-
+from unitcell import unitcell
+from columnfile import columnfile
 """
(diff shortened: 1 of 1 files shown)


ImageD11/peakselect.py
```diff
@@ -73,6 +74,21 @@ def rings_mask(cf, dstol, dsmax, cell=None):
             m |= (abs(cf.ds - v) < dstol)
     return m
 
+def filter_peaks_by_phase(
```
jadball, line 77: This type hinting is unsupported for Python 2.7 and raises a linting error

PR #399, opened 2025-02-13
Initial 3DXRD workflow Specification at /docs/ewoksrst /*.rst and respective .md and README.md at /docs/ewoks_specs

abmajith:
@loichuder


jadball:
@abmajith Do we need both the RST and MD files? Just one or the other should be fine

loichuder:
If there is no strong incentive to use `.rst` (i.e. integrate in an existing doc that uses RST), I would favor MD for its lighter syntax.

abmajith:
Then RST, There are already quite a .rst files are there, so, I will remote the .md files

jonwright:
Merged - peakselect was broken

PR #401, opened 2025-02-14
update on do_index() function in indexing.py file

abmajith:


ImageD11/indexing.py
jadball, line 1: You need to change this to:

```python
    if unitcell:
        indexer = indexer_from_colfile_and_ucell(cf_for_indexing, ucell=unitcell)
    else:
        indexer = indexer_from_colfile(cf_for_indexing)
```

Note that `cf` should be replaced with `cf_for_indexing` as per my comment in the issue you raised.

jadball:
Thanks!

PR #406, opened 2025-02-20
Notebook annoyance fixes

jadball:
Please keep open for a little in case I spot any more before the release!

### DIFF
--- a/ImageD11/unitcell.py
+++ b/ImageD11/unitcell.py
@@ -179,7 +179,7 @@ class Phases(AnalysisSchema):
         phase_name: the name of the phase
         unitcell: the unitcell object
         """
-        super().add_phase_from_unitcell(phase_name, unitcell)
+        super(Phases, self).add_phase_from_unitcell(phase_name, unitcell)
         self.get_unitcells()
     
     def get_unitcells(self):


ImageD11/peakselect.py
```diff
@@ -10,6 +10,103 @@
...
+        self.p1 = None  # Index of first peak
+        self.p2 = None  # Index of second peak
+        self.xy = None  # Fitted position
+        self.peak_ycalc = None  # Computed dty values
+        
+        self._setup_plot()
+        
+    def _setup_plot(self):
+        plt.ion()  # Enable interactive mode
+        self.fig, self.ax = plt.subplots(1, 2, figsize=(10, 7), sharex=True, sharey=True, layout='constrained')
```
jonwright, line 34: This requires a plot at the end of init. It was previously in a GUI subfolder. Right now, it won't run without a GUI, so either it stops plotting, or it should go back into the nbGUI area

ImageD11/sinograms/dataset.py
```diff
@@ -642,16 +647,27 @@ def sinohist(self, weights=None, omega=None, dty=None, method="fast"):
             omega = self.omega_for_bins
         if dty is None:
             dty = self.dty
+            
+        if (self.omega.max() - self.omega.min()) > 360:
+            om_mod = omega % 360
```
jonwright, line 652: This is already computed in self.guess_bins and defined as self.omega_for_bins (used above).

ImageD11/peakselect.py
```diff
@@ -57,6 +57,36 @@ def assign_ring_histo(cf, dsmax, hbins, cell):
     return (bcen,h,ra)
 
 
+def mask_by_ifrac(colf, dstol, dsmax, ifrac, uc, forref=None):
```
jonwright, line 158: This selects rings for a phase (uc) and then masks them. So maybe "mask_rings_by_ifrac"

ImageD11/peakselect.py
jadball: I lifted this from the `select_for_index_unknown` notebook - I'll move it to somewhere in `nbGUI`

ImageD11/peakselect.py
jadball: Done :)

ImageD11/sinograms/dataset.py
jadball: This is the case for `self.omega`, but not for the user-supplied `omega` value, which comes from the `cf`, and therefore may not have been modded yet. I think it's okay?

jonwright: OK

jadball:
Just waiting for FeAu notebooks to finish for the `docs` folder, then I think I'm happy to merge

jadball:
Rejoice! Rejoice! All checks have passed!

jadball:
Merging now, will add comments tomorrow explaining what's in this PR.

PR #408, opened 2025-02-21
Ensure that files are closed after reading/saving parameters

loichuder:


ImageD11/parameters.py
```diff
@@ -278,12 +278,11 @@ def saveparameters(self, filename):
         """
         Write parameters to a file
         """
-        f = open(filename, "w")
-        keys = list(self.parameters.keys())
-        keys.sort()
-        for key in keys:
-            f.write("%s %s\n" % (key, str(self.parameters[key])))
-        f.close()
+        with open(filename, "w") as f:
```
loichuder, line 281: This is the recommended syntax since Python 2.5 to avoid explicit call to `close` and ensure file closing on error: https://docs.python.org/2.7/library/stdtypes.html#file.close

ImageD11/parameters.py
```diff
@@ -295,7 +294,8 @@ def loadparameters(self, filename, phase_name=None):
             pars_dict = AnalysisSchema(filename=filename).get_xfab_pars_dict(phase_name)
             self.parameters.update(pars_dict)
         else:
-            lines = open(filename, "r").readlines()
+            with open(filename, "r") as f:
+                lines = f.readlines()
```
loichuder, line 298: This one was not closed.

jadball:
LGTM, thanks!

Issue #411, opened 2025-02-22
Bug in `indexing.py` - `RuntimeWarning` should be `ValueError`


AxelHenningsson:
Hi!

The bug ([link to bad code](https://github.com/FABLE-3DXRD/ImageD11/blob/6b56317c0176568ad6036cb52ab9f3dac35ada53/ImageD11/indexing.py#L256))
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
There is a bug in [`indexing.py`](https://github.com/FABLE-3DXRD/ImageD11/blob/6b56317c0176568ad6036cb52ab9f3dac35ada53/ImageD11/indexing.py#L1088) in the [`indexer.refine()`](https://github.com/FABLE-3DXRD/ImageD11/blob/6b56317c0176568ad6036cb52ab9f3dac35ada53/ImageD11/indexing.py#L1088). [The following try-except block will not catch the claimed error of an empty `contribs` array:](https://github.com/FABLE-3DXRD/ImageD11/blob/6b56317c0176568ad6036cb52ab9f3dac35ada53/ImageD11/indexing.py#L256) 
 
```python
  contribs = drlv2[ind]
  try:
      fitlastrefined = math.sqrt(np.sum(contribs) / contribs.shape[0])
      if not quiet:
          logging.debug("after %.8f %5d" % (fitlastrefined, contribs.shape[0]))
  except:
      logging.error("\n\n\n")
      logging.error("No contributing reflections for %s\n" % (str(UBI)))
      logging.error("After refinement, it was OK before ???")
      logging.error("\n\n\n")
      return UBI
```

Minimal reproducible example
----------------------------------------------------------------------
```python
import numpy as np
import math

contribs = np.array([])
try:
    fitlastrefined = math.sqrt(np.sum(contribs) / contribs.shape[0])
    #if not quiet:
    #    logging.debug("after %.8f %5d" % (fitlastrefined, contribs.shape[0]))
except:
    #logging.error("\n\n\n")
    #logging.error("No contributing reflections for %s\n" % (str(UBI)))
    #logging.error("After refinement, it was OK before ???")
    #logging.error("\n\n\n")
    #return UBI
    raise
print("I did not enter the except block")
```
output is a `RuntimeWarning` with no entry into the except block:
```
>> I did not enter the except block
>> RuntimeWarning: invalid value encountered in scalar divide
>> fitlastrefined = math.sqrt(np.sum(contribs) / contribs.shape[0])
```

[...]

PR #412, opened 2025-02-22
Bugfix in `indexing.indexer.refine()` #411

AxelHenningsson:
see [#411](https://github.com/FABLE-3DXRD/ImageD11/issues/411#issue-2872711066)

### DIFF
--- a/ImageD11/indexing.py
+++ b/ImageD11/indexing.py
@@ -253,11 +253,11 @@ def refine(UBI, gv, tol, quiet=True):
     ind = np.compress(np.less(drlv2, tol), np.arange(gv.shape[0]))
     # scorelastrefined=ind.shape[0]
     contribs = drlv2[ind]
-    try:
+    if contribs.size != 0:
         fitlastrefined = math.sqrt(np.sum(contribs) / contribs.shape[0])
         if not quiet:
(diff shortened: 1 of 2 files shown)


AxelHenningsson:
This seems to be the case also [here](https://github.com/FABLE-3DXRD/ImageD11/blob/6b56317c0176568ad6036cb52ab9f3dac35ada53/ImageD11/indexing.py#L1146)

jadball:
@AxelHenningsson nice catch! Can you please make the associated change, to catch the `ValueError` too in the `except` block below? I fear raising the exception will break the point-by-point indexing unless we catch it suitably

https://github.com/FABLE-3DXRD/ImageD11/blob/6b56317c0176568ad6036cb52ab9f3dac35ada53/ImageD11/sinograms/point_by_point.py#L1803-L1809

AxelHenningsson:
[Like this James?](https://github.com/FABLE-3DXRD/ImageD11/pull/412/commits/a1b4dd2a0079bbb71add3d9f1eee918af31fbd35)

jadball:
Yup! Thanks, I'm happy with this but I think we should wait for @jonwright just in case before merging :)

jonwright (approved):
Thanks!

PR #418, opened 2025-02-25
update on indexer_from_colfile_and_ucell function

abmajith:
@jonwright, @jadball, @loichuder 
even if wavelength is present in kwds, Python evaluates float(colfile.parameters.get("wavelength")) before calling kwds.get().

This happens for some reason, and produces an error, so I had to rewrite the logic like this 

```
    w = kwds.get("wavelength")
    if w is None:
        w = colfile.parameters.get("wavelength")
        if w is None:
            raise ValueError("Wavelength must be provided either in kwds or colfile.parameters")
```


jadball:
looks good, thanks!

Issue #384, opened 2025-01-21
colfile_to_hdf: TypeError: Only chunked datasets can be resized


jadball:
We see this error a lot with `columnfile.colfile_to_hdf`, it would be nice if it was handled more gracefully with a try/except...

```
E           File (...)/ImageD11/ImageD11/columnfile.py:591, in colfile_to_hdf(colfile, hdffile, name, compression, compression_opts)
E               589 if t in list(g.keys()):
E               590     if g[t].shape != dat.shape:
E           --> 591         g[t].resize( dat.shape )
E               592     g[t][:] = dat
E               593 else:
E           
E           File ()...)/lib/python3.11/site-packages/h5py/_hl/dataset.py:666, in Dataset.resize(self, size, axis)
E               664 with phil:
E               665     if self.chunks is None:
E           --> 666         raise TypeError("Only chunked datasets can be resized")
E               668     if axis is not None:
E               669         if not (axis >=0 and axis < self.id.rank):
E           
E           TypeError: Only chunked datasets can be resized
```

jadball:
The underlying issue:

When you call `colfile_to_hdf` on a columnfile in RAM, and there's a columnfile already existing on disk at `hdffile`, what should the behaviour be?
Right now, if the columnfiles have the same number of `nrows`, then the default behaviour is to overwrite the file on disk (e.g. adding new columns or recomputing some columns).
If the number of `nrows` differs, then we get the error above.

- What should the default behaviour be?
- Should we introduce an option to `raise` upon collision? Perhaps with a simple `os.path.exists`?

jonwright:
Today we over-write, so continuing to do this seems OK to me? In general `excel -> save spreadsheet` over-writes your file, so this is not unexpected behaviour for a newcomer. But maybe people want to save an "undo history" to go back later? For the text files today, I think they over-write.

If we want to write-protect files to avoid accidental trashing then we should do that with file permissions within #373 (for example, assembly of the many sparse file to use links rather than copies, and then we don't want to delete the linked files).

This is more of a technical problem with hdf5. Options include:
- change the error message to tell the user to delete the file if this is really what they want
- we can delete the dataset and replace by a new one if the size does not match and it is not chunked. This does not reclaim the space in the file unless we also call h5repack.
- write a new file and then copy over anything which was not part of our columfile from the existing h5 (equivalent to the last step, but harder to write)
[...]

Issue #413, opened 2025-02-24
Inconsistent Parameter Saving Between .flt and .h5


abmajith:
HI @jonwright , @jadball 

In the long term, we plan to support saving a list of parameters in both .flt and .h5 formats. 
Ideally, these lists of parameters should remain consistent across both formats.

Currently, I noticed a difference in how parameters are saved:

    The columnfile class's writefile method saves all attributes.
    The colfile_to_hdf function mostly (or only) saves title attributes.

This causes an issue during indexing, where it expects to read the wavelength attribute from the loaded columnfile (from the .h5 file), but it isn't available, while using the function 
```
venv/lib/python3.12/site-packages/ImageD11/indexing.py:292: in indexer_from_colfile_and_ucell
    w = float(colfile.parameters.get("wavelength"))
```

Would it be sufficient to work around this issue by relying solely on the .flt format using writefile and readfile methods?
Or do we need to extend the .h5 saving method to ensure it stores all attributes, similar to how .flt does?

jonwright:
Hi @abmajith 

It is not very clear to me right now. A columnfile can also be coming from an mmap.

Historically we intended to save in the hdf5:  https://github.com/FABLE-3DXRD/ImageD11/blob/5ead15d915176cdea0be73bcd4a4e3b3387b445d/ImageD11/columnfile.py#L559

I suspect you need parameters for each column in the dataset. For example, the "raw data" columns would refer to the peaksearch threshold and frames. The sc/fc would refer to the spline file. The tth/eta/gx/gy/gz would refer to the geometry. The 'ring_id' would refer to a unit cell.

It might have been better if we had not saved parameters into colfiles, but we cannot change the past.

loichuder:
Related to this, is the HDF5 saving/loading of ColumnFiles used somewhere ?

We are currently considering different ideas to save CF as NeXus-compliant data and we could implement this directly inside the ColumnFile class (either as a replacement or as a different HDF5 saving/loading) if you are interested ?

jonwright:
Yes - columnfiles are frequently loaded and saved as hdf in the notebooks. We sometimes use the memory mapped columnfiles backed by hdf5 (in point by point indexing I think).

If you can make a Nexus compliant version it would be great. Can it be made compatible by just adding some attributes?

PR #420, opened 2025-02-27
Ewoks task specification review

jadball:
@abmajith @loichuder Here's my review of the Ewoks task specification. I've moved folders and renamed a few tasks for clarity.

# SegmentFrame
I recommend not limiting the choice of `omegamotor` or `dtymotor` - this may change at random if a detector gets swapped out or renamed in BLISS etc. I changed the documentation to provide examples rather than restricting the choice.

> **Question**
> 
> Is there a clever way of dumbing all the configuration of *segmenter parameters*, *folder file* settings inside the 
> spatial_corrected_3d_file.h5 or segmented_3d_file.h5 without affecting the columnfile usage inside ImageD11?

I see you opened an issue about this - #413 
I suspect that this isn't an easy problem to solve with the existing implementation of `H5` columnfiles. Is this issue blocking for you, or could you come up with your own method to solve this?
I guess it depends on whether you absolutely want the parameters saved in the `.h5` columnfile, or whether you'd be happy with a separate file?
Let me know - if it is essential for you, I can try to implement parameter saving in H5 files and see if it breaks anything.


abmajith:
> @abmajith @loichuder Here's my review of the Ewoks task specification. I've moved folders and renamed a few tasks for clarity.
> # SegmentFrame
> 
> I recommend not limiting the choice of `omegamotor` or `dtymotor` - this may change at random if a detector gets swapped out or renamed in BLISS etc. I changed the documentation to provide examples rather than restricting the choice.

Hi @jadball 
Yes, it should be simple string, not from enumerated string list,  In ID03, it specified omegamotor as omega, not diffrz as in ID11. 


> 
> > **Question**
> > Is there a clever way of dumbing all the configuration of _segmenter parameters_, _folder file_ settings inside the
> > spatial_corrected_3d_file.h5 or segmented_3d_file.h5 without affecting the columnfile usage inside ImageD11?

I think, more simpler way is maintaining folder for each phase string like Fe, Au, and store the respective .par, filtered peaks, grains etc. 
But it should be discussed with @loichuder , what he thinks about it

For me the Ideal way is maintaining a folders like Analysis_folder/{sample}/{sample}_{dataset}/Fe_{tag1}, Fe_{tag2}, Au_{tag3}, Au_{tag4}, represents the experimented phase/lattice name along with its parameter 

In this case, I dont have to append any new key value to the .h5 file, or do changes on ImageD11 side.

abmajith:
> I see you opened an issue about this - #413 I suspect that this isn't an easy problem to solve with the existing implementation of `H5` columnfiles. Is this issue blocking for you, or could you come up with your own method to solve this? I guess it depends on whether you absolutely want the parameters saved in the `.h5` columnfile, or whether you'd be happy with a separate file? Let me know - if it is essential for you, I can try to implement parameter saving in H5 files and see if it breaks anything.

Actually, it is not blocking, we are not rushing to one quick solution,

After getting some information from columnfile, we could add all possible key value (or attribute and its value) without blocking other application. 

In this way, along with key, value pairs in .h5, we could also add the user inputs likes tolerence, ds.max(), lattice_name, lattice_parameters and space group information. 

Will do one thing, I will try and see how its coming along, and after I will create a merge request.

jonwright:
@loichuder : when using ImageD11, a columnfile can add new columns and/or overwrite existing values and then get saved back into the same file. They are used like excel spreadsheets (open->edit->save).

I assume we will fix #384, so then people can delete rows out of them, which was always the default and allowed for text files.

They have no notion of "history" or "undo" or how they are created. So it would be great if ewoks handles things arising like that.

loichuder:
> @loichuder : when using ImageD11, a columnfile can add new columns and/or overwrite existing values and then get saved back into the same file. They are used like excel spreadsheets (open->edit->save).

I see. But you do this in specific groups right, like `peaks` ? These have a bunch of 1D datasets inside representing the columns in question.

```
root
├── peaks
    ├── ds
    ├── eta
    └── ...
```

My question is would it break anything to add other groups alongside these ? Something like


```
root
├── peaks
│  ├── ds
│  ├── eta
│  └── ...
└── my_new_grp
           ├── some stuff
           ├── not columns
           └── ...
```

jonwright:
Should be no problem. ImageD11 columnfile is only meant to look at groups having `group.attrs["ImageD11_type"]="peaks"`. It will get confused if there are several peaks groups, but you can tell it which one to load ( eg: `colfile_from_hdf( hdffile , name='/path/to/your/data'`). It should not worry about anything else in the file.

loichuder:
Good to know, thanks :slightly_smiling_face:

PR #426, opened 2025-02-28
I added two methods in columnfile class of columnfile .py file

abmajith:
Hi @jonwright, @jadball , et @loichuder,

write_hdf5 method explicitly create two groups, one groups related to the parameter.parameter() dict, 

another group is to store the titles and its big array.

and loading the columnfile class instance from given hdf5 is written in read_hdf5() as class method.

By doing this two groups, the ewoks team can create additional group for stroing the filtering parameter, or used lattice parameters etc.

### DIFF
--- a/ImageD11/columnfile.py
+++ b/ImageD11/columnfile.py
@@ -285,6 +285,15 @@ class columnfile(object):
             fout.write(format_str % tuple( [col[i] for col in self.__data] ) )
         fout.close()
 
+    def write_hdf5(self, filename):
+        with h5py.File(filename, "w") as f:
+            meta_group = f.create_group("parameters")
+            for key, value in self.parameters.get_parameters().items():
+                meta_group.attrs[key] = value
+            data_group = f.create_group("columndata")
+            for title, col in zip(self.titles, self.bigarray):
+                data_group.create_dataset(title, data=col)
+
     def readfile(self, filename):
         """
         Reads in an ascii columned file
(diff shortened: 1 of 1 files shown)


jonwright:
Hi @abmajith,
Can you look at the existing `colfile_from_hdf` and `colfile_to_hdf` methods that are further down in the file and not part of the class definition? Then try to figure out how to merge your code with the existing approach. Perhaps something like:

hdf5file:
```
  /path/to/peaks_table  -> { dataset( name , data=colf['name'] ) for name in colf.titles }
      /path/to/peaks_table.attrs[ 'ImageD11_type'] = 'peaks'
      /path/to/peaks_table.attrs[ 'ImageD11_parfile_link'] =  '/path/to/peaks_table_parameters' 
  /path/to/peaks_table_parameters -> however you want to store them
      /path/to/peaks_table.attrs[ 'ImageD11_type'] = 'parameters'
```
If you want to store Nexus parameters then I guess the way to store parameters will be different.

jadball:
I agree with @jonwright - I think we shouldn't add new methods to the `columnfile` class right now. This is maybe the most fundamental class for ImageD11 and it's likely to break many workflows if we change it. Another approach is for you guys to handle this in the Ewoks package rather than pushing this upstream @abmajith

abmajith:
> I agree with @jonwright - I think we shouldn't add new methods to the `columnfile` class right now. This is maybe the most fundamental class for ImageD11 and it's likely to break many workflows if we change it. Another approach is for you guys to handle this in the Ewoks package rather than pushing this upstream @abmajith

Yes, Okay, will do one thing, these columnfiles titles and parameters attributes (key, values) are straight forward  (as of now for me), I will create one nexus format and will refine it with Wout, Loic and of course ID11 teams, , and we will keep these hdf5 nexus write and read within ewoks3dxrd.

jadball:
Thanks @abmajith - I will close this for now.

PR #429, opened 2025-03-03
f2scan blocked series by HF

jonwright:


jadball:
LGTM, will merge in

haixing0a:
@abmajith, since yesterday @jonwright and @jadball made the code to handle f2scan for s3DXRD and the tests were successful. Maybe we overlooked it for the box-beam 3DXRD. @abmajith, I copy your comments below so that @jonwright and @jadball can follow:


"""
Hi @haixing0a This update actually break the existing masterfile read, for me I am getting this
>   self.omega = np.array(self.omega).reshape(self.shape)
E       ValueError: cannot reshape array of size 1440 into shape (0,0)

venv/lib/python3.12/site-packages/ImageD11/sinograms/dataset.py:487: ValueError

this is happened
as

venv/lib/python3.12/site-packages/ImageD11/sinograms/dataset.py:254: in import_all
    self.guess_shape()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = dataroot = "/mnt/id03_3dxrd/expt/RAW_DATA"
analysisroot = "/tmp/pytest-of-noordhee/pytest-6/test_segmentation_workflow0/test_my_task"
sample = "FeAu_0p5_tR"
dset = "ff1"
shape = ( 0, 0)
# scans 1 from 1.1 to 1.1

"""

jonwright:
@abmajith Is there an example we can look at?

abmajith:
> @abmajith Is there an example we can look at?

Hi @jadball, @jonwright and @haixing0a 


the problem is exactly here 

```
if len(self.scans) > 1:
            s0 = len(self.scans)
            s1 = npts // s0
else:
            # no scans
            s0 = 0
            s1 = 0
```

originally it is intended to like this

if len() == 1: will load s0, s1
elif len()  > 1: will load s0,s1 appropriately
else: i.e len() == 0 will put s0,s1 to zero

 after the logic was rewritten, the logic for codingion len() == 1 goes to else: i.e len() == 0 will put s0,s1 to zero

jonwright:
Ah! It seems we miss a testcase ... you want that dset.scans = (1, nframes ) ?

jonwright:
So I sent a pull #432 - for the testcase I guess we need to look at making one

PR #434, opened 2025-03-05
Fix error when saving grain with no translation

loichuder:
`grain.translation` can be `None`: https://github.com/FABLE-3DXRD/ImageD11/blob/0daa4f1e629a47de707aa7ce0344e1d09ae83cdc/ImageD11/grain.py#L55

### DIFF
--- a/ImageD11/grain.py
+++ b/ImageD11/grain.py
@@ -316,7 +316,8 @@ def write_grain_file(filename, list_of_grains):
     f = open(filename, "w")
     for g in list_of_grains:
         t = g.translation
-        f.write("#translation: %g %g %g\n" % (t[0], t[1], t[2]))
+        if t is not None:
+            f.write("#translation: %g %g %g\n" % (t[0], t[1], t[2]))
         if hasattr(g, "name"):
             f.write("#name %s\n" % (g.name.rstrip()))
         if hasattr(g, "intensity_info"):


jadball:
This seems okay to me... @jonwright ?
In the `read_grain_file` method we only apply a non-`None` translation if we find one in the file:

```python
def read_grain_file(filename):
    """read ubifile and return a list of ubi arrays """
    f = open(filename, "r")
    grainsread = []
    u = []
    t = None   ### <- HERE - it's None by default
    p = {}
    for line in f:
        if line.find("#translation:") == 0:
            t = [float(x) for x in line.split()[1:]]  # only changed from None if we find a translation in the file
            continue
        if line[0] == "#" and line.find("UBI") < 0:
            k, v = line[1:].split(" ", 1)
            p[k] = v
            continue
        if line[0] == "#" and line.find("intensity_info") > -1:
            p["intensity_info"] = line.split("intensity_info")[1].rstrip()
        if line.find("#") == 0: continue
        vals = [float(x) for x in line.split()]
        if len(vals) == 3:
            u = u + [vals]
        if len(u) == 3:
            grainsread.append(grain(u, t))
            for k in ["name", "npks", "nuniq", "intensity_info"]:  # Rod - is recomputed when needed
                if k in p:
                    setattr(grainsread[-1], k, p[k])
            p = {}
            u = []
            t = None
    f.close()
    return grainsread
```

PR #435, opened 2025-03-05
Add option to scale intensities

jadball:
## Eiger data

To automatically scale the peak intensities by a monitor signal, we set an active monitor to the dataset class.

To normalise to a monitor signal, we need to choose a "reference" monitor value that we scale to.
A good choice may be `np.mean(monitor_per_frame)`

We then compute `scale_factor_per_frame = np.mean(monitor_per_frame) / monitor_per_frame`  
We then multiply the observed intensities by `scale_factor_per_frame`

You can choose which function to use to generate the "reference" monitor value, for example:

`ds.set_monitor(monitor_name='fpico6', ref_value_func=np.mean)`

Once a `monitor` has been set, it should be applied by default whenever a `cf_2d` or `cf_4d` is auto-computed from the `peaks_table`. Note this is only relevant to outputs of `ImageD11.sinograms.properties`.

## Frelon data

In a similar way, when you choose to run `segment_dataset`, you can pass an optional `monitor_name` and `monitor_ref_func`:

```python
cf_2d, cf_3d = segment_dataset(ds, options, monitor_name=monitor_name, monitor_ref_func=np.mean)
```

**This correction applies to the raw frame data before performing the background subtraction.**  
@jonwright are you happy with this?

```python
def peaksearch(self, img, omega=0, scale_factor=None):
    if self.wrk is None:
        self.wrk = np.empty(img.shape, "b")
        self.labels = np.empty(img.shape, "i")
    if scale_factor is not None:
        # scale this image by the scale factor
        img = img * scale_factor
    
    self.cor = self.bgsub(img)
    self.cor = self.masksub(self.cor)
    # smooth the image for labelling (removes noise maxima)
    self.smoothed = scipy.ndimage.gaussian_filter(self.cor, self.smoothsigma)
```

@abmajith , if this PR is accepted, it may mean another task for you to consider whether to correct by intensities - maybe alongside a new task that previews the monitor column to see whether you want to apply it.
See the example notebook in `ImageD11/nbGui/TDXRD/0_segment_frelon.ipynb` for example usage.

 It's worth checking with ID03 whether they have appropriate in-beam intensity monitors (we have a few pico monitors at different points along the beam path we can choose to insert or not). Currently, peak intensities are not really considered as part of the standard box-beam indexing process, although this may change in the (near or far) future.


jonwright:
If you want to scale the image you would need do it after the dark/flat corrections. The code above will rescale the dark and that create problems.

ImageD11/frelon_peaksearch.py
```diff
@@ -339,11 +339,14 @@ def masksub(self, img):
             img_masked[self.mask == 1] = 0
         return img_masked
 
-    def peaksearch(self, img, omega=0):
+    def peaksearch(self, img, omega=0, scale_factor=None):
         if self.wrk is None:
             self.wrk = np.empty(img.shape, "b")
             self.labels = np.empty(img.shape, "i")
-
+        if scale_factor is not None:
```
jonwright, line 346: You need to pass the scale_factor into bgsub and apply it after self.correct(img).

jonwright:
Given that frelon people are going to have to re-process their data I will add a "todo" on the sparse pixel refactoring to try to read labels and intensities from the frelon code.

jadball:
@jonwright like this? https://github.com/FABLE-3DXRD/ImageD11/pull/435/commits/0b742c9be46a4e44b7ad5e0e77b9d2670a4d984d

abmajith:
Hi @jadball 

In the file "frelon_peaksearch.py", function 
```
def segment_master_file(
    master_file,
    frames_dataset,
    omega_angles,
    worker_args,
    num_cpus=None,
    scale_factor=None,
    **process_map_kwargs
):
    """
    Collects peaks for all the frames in the first scan in the dataset 
    using parallel processing.
    """
    num_threads = num_cpus or max(1, ImageD11.cImageD11.cores_available() - 1)
    num_frames = omega_angles.shape[0]
    args = [
        (master_file, frames_dataset, i, omega_angles[i], worker_args, scale_factor[i]) 
        for i in range(num_frames)
    ]
    all_frames_peaks_list = process_map(
        pps, 
        args, 
        chunksize=1, 
        max_workers=num_threads, 
        **process_map_kwargs
    )
    return all_frames_peaks_list
```

for the default input i.e scale_factor = None, this function will throw an  TypeError
```
E       TypeError: 'NoneType' object is not subscriptable
venv/lib/python3.12/site-packages/ImageD11/frelon_peaksearch.py:492: TypeError
```

jadball:
> for the default input i.e scale_factor = None, this function will throw an TypeError
> And also, it should be asserted that length of omega_angles and scale_factor should be same.

@abmajith good spot - I will open a new PR to fix this.

> Is it possible to grab the omega_angles and this scale_factor from master file itself rather than providing as a separate arguments?
> In dataset Class,monitor, monitor_ref, omega_angles are derived from master file, okay for this it needs scan_number

It's technically possible, but that's why we have the `segment_dataset` function, which will determine the appropriate H5 paths to the frames as well as correctly reading the monitor and omega angles.

abmajith:
> > Is it possible to grab the omega_angles and this scale_factor from master file itself rather than providing as a separate arguments?
> > In dataset Class,monitor, monitor_ref, omega_angles are derived from master file, okay for this it needs scan_number
> 
> It's technically possible, but that's why we have the `segment_dataset` function, which will determine the appropriate H5 paths to the frames as well as correctly reading the monitor and omega angles.


Yes, In our Ewoks task, we are handling the segmentation directly from masterfile, okay, we can grab 
omega values, monitor ref and etc before calling segment_master_file functions 


We also contacted ID03 regarding the Monitor intensities  normalization, they do have such pico4, pico3 monitors, but their instrument it was called in 1.2 (i.e ASYNC), so have to come up with interpolation before using the intensity normalization as by your code.

PR #438, opened 2025-03-06
Fix error in segment_master_file when scale_factor is None

loichuder:
```
    args = [
>       (master_file, frames_dataset, i, omega_angles[i], worker_args, scale_factor[i])
        for i in range(num_frames)
    ]
E   TypeError: 'NoneType' object is not subscriptable
```


loichuder:
Actually already fixed by https://github.com/FABLE-3DXRD/ImageD11/pull/437.

Sorry for the noise.

PR #444, opened 2025-03-11
fixes #269

### LINKED ISSUE #269: sinograms.geometry module
Handles things like:

- [x] iy index to/from dty value (accounting for y0 and ystep)
- [x] recon array indices to/from real space (x,y) values (accounting for y0, ystep and padding)
- [x] (dty,omega) values to/from real space (x,y) values (via fitting sine waves)
- [ ] test/debug whether sinograms/roi_iradon.py:recon_cens matches the same geometry

Also needs unit tests

jonwright:


jonwright:
This should fix + demo the recon_cens code

Issue #291, opened 2024-06-20
Allow no mask to be used with lima_segmenter


jonwright:
```
  File "/cvmfs/hpc.esrf.fr/software/packages/linux/x86_64/jupyter-slurm/2023.10.7/envs/jupyter-slurm/lib/python3.11/site-packages/ImageD11/sinograms/lima_segmenter.py", line 308, in segment_lima
    for i, spf in enumerate(reader(frms, OPTIONS.mask, OPTIONS.cut)):
                                         ^^^^^^^^^^^^
AttributeError: 'SegmenterOptions' object has no attribute 'mask'
```

jonwright:
Done in #445

PR #445, opened 2025-03-11
Allow no mask (needs testing)

jonwright:


jonwright:
Would close #291 if it works

PR #453, opened 2025-03-17
dataset.obincens

jonwright:
Please do not merge yet. 

ds.ostep is still wrong.

We have:
```
1.1 : fast_start_pos = -119.55, fast_step_size = 0.05, fast_npoints = 7220
2.1 : fast_start_pos =  263.15, fast_step_size = 0.05, fast_npoints = 7220
...
10.1 : fast_start_pos =  3324.75, fast_step_size = 0.05, fast_npoints = 7220
```

We could pick up ds.ostep from master -> scan/instrument/fscan_parameters. 

Or we could leave it as a manual fix for this kind of case.

Or we could fit the slope omega vs frame number slope (interlaced?) for each row of the sinogram.

### DIFF
--- a/ImageD11/sinograms/dataset.py
+++ b/ImageD11/sinograms/dataset.py
@@ -38,6 +38,36 @@ def guess_chunks(name, shape):
     return shape
 
 
+def guess_omega_step( omega, rptcut=0.02 ):
+    """
+    Estimates the step in omega angle data for multi-turn
+
+    Case 1: more turns try to interleave intentionally (irrationals)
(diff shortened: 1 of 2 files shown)


jonwright:
This tries to fix things for Wenxi
- multiprocessing glitch (sem lock issue with many processes starting together on the same machine)
- guessing a sensible omega step from the data

Starts to look like it will never be debugged/correct.

jonwright:
So I will merge in order to do the release sooner rather than later.

PR #454, opened 2025-03-18
Fix wrong frelon flat/dark correction

jadball:
It seems flat was overwriting dark rather than chaining the correction?


jonwright:
Yes - it looks like it was wrong.

PR #459, opened 2025-03-20
fix memory issues for large cf comparison in forward_model etc

haixing0a:
fix memory issues for large cf comparison in forward_model, correct typos and add a few print outputs. It should be the last PR from my side before the new release.


haixing0a:
> Thanks! I will try to upload another one to pypi tomorrow - probably with power9 turned on again

Thanks ! I will merge this in now.

PR #470, opened 2025-03-31
Friedel pair position mapping

jonwright:
New notebook to locate peaks inside the sample from scanning data.


jonwright:
flake8 has been updated: see https://github.com/sympy/sympy/issues/27855

PR #472, opened 2025-04-02
Fix #471

### LINKED ISSUE #471: refine_map numba function is segfaulting
This seems to be a regression. `refine_map` is now segfaulting.
Seems to be sensitive to the number of points being simultaneously refined:

`refine.run_refine(use_cluster=False, npoints=119)` runs okay
`refine.run_refine(use_cluster=False, npoints=120)` segfaults

jadball:
Fix #471


jonwright:
Ignore the pylint ci error for now. They should be fixable and 3.8 was not broken.

PR #475, opened 2025-04-08
Fix setup.py with CFLAGS/LDFLAGS

t20100:
This PR:
- fixes a compilation issue if there is 2 spaces in CFLAGS or LDFLAGS env. variables (occuring in new jupyter-slurm conda env). In this case an empty string is passed as an argument to the compiler as `""`.
- avoids the use of `self.extensions` and use `ext` instead since it is the currently handled Extension
- avoid use of `ext_package` and use the fully qualified name in the Extension: Without this, the `_cImageD11` is advertised as a "top level" module (I don't know why, it shouldn't). To check: `from importlib.metadata import packages_distributions; packages_distributions()['_cImageD11']`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,9 +78,9 @@ class build_ext_subclass( build_ext.build_ext ):
         c = self.compiler.compiler_type
         CF = [] ; LF=[]
         if "CFLAGS" in os.environ:
-            CF = os.environ.get("CFLAGS").split(" ")
+            CF = os.environ.get("CFLAGS").split()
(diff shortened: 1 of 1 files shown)


t20100:
BTW, I don't think adding `CFLAGS` and `LDFLAGS` is needed: Those flags looks to be duplicated in the compiler calls. `setuptools` looks to already do the job.

jonwright (approved):
Looks good, thanks!

PR #488, opened 2025-05-16
bugfix forfrelon segmenter when intensity normalization is active

AxelHenningsson:
Zombie code has awaken! :zombie: 

When iterating multiple scans for frelon segmentation with intensity normalisation active, there is a bug :bug:  in the indexation that causes a hard crash regradless of input. The typo has been fixed.

This implies that the template `0_segment_frelon.ipynb` is broken before this patch. With respet to this template, please also note that some of the defult frelon files (such as the dark) is not setup for user permission :see_no_evil: 

Cheers
Axel


jonwright:
Merging as it looks right.
What is the file permissions problem?

PR #489, opened 2025-06-03
Add the capability to write xdmf text for 5-D data in the function Te…

haixing0a:
Add the capability to write xdmf text for 5-D data (including eps_sample, U, B, UBI, UB and mt) in the function TensorMap.to_paraview. It is appealing to be able to visualize all the fields of the tensor map within ParaView, especially for eps_sample. Tested with user experiment es1603. @jadball and @jonwright please review this update.


jonwright:
Looks good to me - leaving it to @jadball to merge as he wrote the tensormap code

PR #501, opened 2025-07-03
Update release workflow

### LINKED ISSUE #500: release builds are slow
Some maintenance is needed in the build_and_deploy workflow as it is compiling dependencies. If users are installing binary wheel for ImageD11, they should be doing that for the dependencies too). To be checked with @t20100

t20100:
This PR updates the release workflow to use `numpy` wheels for all platforms.

It also remove `six` from build dependencies since it is apparently not used for building the package.

closes #500

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
           platforms: all
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.16.5
+        uses: pypa/cibuildwheel@v3.0.0
         env:
           CIBW_BUILD_VERBOSITY: 1
           CIBW_PROJECT_REQUIRES_PYTHON: ">=3.8"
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -27,7 +27,7 @@ jobs:
             cibw_archs: "auto64"
           #- os: windows-2019
           #  cibw_archs: "auto32"
-          - os: macos-13
```
t20100, line 30: This is to run on `arm65` machines

.github/workflows/release.yml
```diff
@@ -40,23 +40,27 @@ jobs:
           platforms: all
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.16.5
```
t20100, line 43: This updates the default `manylinux` target, but anyway produced wheels should be compatible with `manylinux` versions down to `manylinux_2014`

.github/workflows/release.yml
```diff
@@ -40,23 +40,27 @@ jobs:
...
+          # Do not build for muslinux and python>=3.12 on ppc64le
+          CIBW_SKIP: "*-musllinux_* cp312-*linux_ppc64le cp313-*linux_ppc64le"
           CIBW_ARCHS: ${{ matrix.cibw_archs }}
 
           # Use silx wheelhouse: needed for ppc64le
           CIBW_ENVIRONMENT_LINUX: "PIP_FIND_LINKS=https://www.silx.org/pub/wheelhouse/ PIP_TRUSTED_HOST=www.silx.org"
 
+          # Build without isolation needed to use older wheels for ppc64le
+          CIBW_BEFORE_BUILD: pip install setuptools wheel "numpy; python_version >= '3.9'" "oldest-supported-numpy; python_version < '3.9'" --only-binary numpy
+          CIBW_BUILD_FRONTEND: "build; args: --no-isolation"
```
t20100, line 57: Pre-install build dependencies to force using a wheel for `numpy`

PR #507, opened 2025-09-15
add convert fsparse to raw image (useful when raw data went to tape) …

haixing0a:
…and other small updates


jonwright:
Sorry - I didn't see this - you also have something like this in ImageD11.sparseframe.SparseFrame.todense()

PR #511, opened 2025-10-08
Fix import_motors_from_master always choosing scan 0

jadball:
Fix `import_motors_from_master` always choosing scan 0 - I believe this is a bug that doesn't normally come up, but does for some 2d scan types. Saw this with much older S3DXRD datasets

### DIFF
--- a/ImageD11/sinograms/dataset.py
+++ b/ImageD11/sinograms/dataset.py
@@ -448,12 +448,16 @@ class DataSet:
                     self.dty[i] = np.full(self.frames_per_scan[i], dty[0])
         for b in bad:
             dom = [
-                (abs(self.omega[i][0] - self.omega[b]), i)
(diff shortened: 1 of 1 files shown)


jonwright:
Looks right. We miss a test case perhaps? The problem was:

scan 1.1 : omega OK from 0-180
scan 2.1 : omega OK from 180-0
scan 3.1 : omega OK from 0-10, then corrupted later, needs to copy scan 0
scan 4.1 : omega OK from 180-170, then corrupted later, needs to copy scan 1

For f2scan or fscan2d, I think we would have to read instrument/fscan_parameters to guess the idealised omega positions. This is only for backtracking over a list of 1D scans (even when collected with fscan2d, us usually do one rotation only).

jadball:
Thanks, agreed :)

PR #522, opened 2025-10-30
moved HNAME from global to module scirpt execution alu segment eiger

abmajith:
@jonwright  Here in the module segment_ma4752, I moved the HNAME global variable under the module script execution (if __main__ block)

and also changed the title from bytes type to string type in segment_scans function in the same module


jonwright:
Code in sandbox is not supported/maintained. I don't see the logic to patch it like this with the hard wired paths left in place?
Regardless, I have a feeling that moving the global HNAME to be defined in the __main__ scope would fail under some versions of processpool (spawn, fork) that are now the default, but I have not tested it. Have you?

PR #576, opened 2026-04-23
KAM + gb misorientation parallel

CiolJegou:
I added 2 main functions for Misorientation computation. 
Both functions are very similar in how they are used: pick a method, get pairs of voxels and their neighbors for a given mask, compute the misorientation in parallel for every pair. 
Both methods can be used with Xfab.symmetry.Umis, or with the disorientation function. Results are similar for cubic (I did not test other symmetries yet). Umis computation time is ~10 times faster.

1. DS_KAM_parallel: KAM computation in parallel with a threshold.
2. DS_misori_parallel: Grain boundary misorientation in parallel (pixel by pixel) --> the misorientation can change along a grain boundary

Sub functions : 

- get_boundary_voxels : Get a mask of the grain boundaries based on the grain labels
- find_neighbors : Get the coordinates of all neighbors of a voxel. The neighbors are defined by a kernel
- KAM_kernel : Create a KxKxK kernel with a gaussian function to pick out neigbors for KAM and GB

### DIFF
--- a/ImageD11/forward_model/grainmaps.py
+++ b/ImageD11/forward_model/grainmaps.py
@@ -1432,7 +1432,7 @@ def DS_KAM_parallel(DS, kernel_size=5, kernel_cutoff=50, crystal_system='cubic',
                 return (k, i, j, angle_mean, coo_neighbors)
             else:
                 return (k, i, j, fill_value, coo_neighbors)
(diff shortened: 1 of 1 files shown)


haixing0a:
Looks good, thanks !

PR #591, opened 2026-07-28
Read unregular sinograms when using ds.import_all() - #590

CiolJegou:
There were 2 problems to address with the current status of the beamline ID11.
Assuming a scan with a specified N_dty horizontal steps and R_step angular step from 0 to 360:

    If 360/R_step= N_r is an integer: the diffractometer will do N_r frames per rotation at some N_dty steps, and N_R+1 frames at some others. Besides, the last rotation will be incomplete. It is possible to correct this problem on the hardware side, by applying a correction factor on the angular step before a scan.

    If 360/R_step= N_r is not an integer: then the number of N_r frames per rotation will also be equal to N_r+1 at certain frames. It is not possible to correct this on the hardware side.

This update allows reading the actual angular position of the motor to check the real number of frames for one full rotation --> function get_rotations_images().
Then, to create the regular grid (needed for all subsequent processing):

    if there is +1 frame in the rotation, it does not include it in the list of scans to process.
    it disregards the last incomplete dty step

This code does not change anything for any regular processing of old data, or regular-shaped data.

### DIFF
--- a/ImageD11/sinograms/dataset.py
+++ b/ImageD11/sinograms/dataset.py
@@ -568,7 +568,7 @@ class DataSet:
                         s0 = 1
                         s1 = npts
                         rotations.append( scan )
-            print(f'len rotation: {len(rotations)}')
+            print('len rotation: ', len(rotations))
             self.scans = rotations
         if len(self.scans) >= 1:
             s0 = len(self.scans)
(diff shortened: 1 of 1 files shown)


jonwright:
Looks promising, thanks!

@jadball do you have test cases for this?

@CiolJegou : is there a way to guess the scan direction (+ or -) and have it still work? Something like `sign = np.mean(np.sign(diffs))`. Then search for the sign opposite to the usual one?

We can merge already, but I guess we need to edit the same file for corrupted musst data and the sign fix too.

jadball:
I remember some recent (post-upgrade) scans that had this issue. I was helping Loic remotely I think to fix it? so I can dig them out and test this solution. I recommend merging for now

Issue #21, opened 2018-05-17
reset indexing in the tk gui without restarting program


jonwright:

jonwright:
Could replace the indexer object, load the gve and parameters

jonwright:
For now - just stop and start the program

PR #592, opened 2026-07-28
dataset y0 caching, structure factor corrections, speedups and bug fixes

jadball:
## `blobcorrector.py`

- New function: `get_corrector`
You can supply a detector h5, e2dx/y or a spline, and it will return a corrector function in decreasing priority:
1. detectorh5
2. e2dx/y
3. spline
4. None (if none were supplied)

This means that regardless of correction type, you always get a corrector function like this:
```python
cf = columnfile(blah blah blah)
corrector = get_corrector(h5file='det.h5')
corrector(cf)  # adds sc, fc in-place
```

## `dataset.py`

- Guess (from mean of ymin and ymax) and cache `y0` to disk
I also updated the notebooks to match
- Add method to return spatial corrector depending on which correction we use - useful when building sinograms in tomo_2 method later.

## `geometry.py`

- Now use optional supplied `y0` (defaults `0.0`) as the initial guess when you fit `sx/y` position to `dty/omega` data

## `tensor_map.py`
- Can now handle TensorMaps of different sizes when Z-stacking or phase-combining. Useful if your `y0` fits (e.g. from tomo_2) give slightly different results which leads to slightly different pads, therefore reconstructions have different shapes
- Optionally don't normalise grain reconstructions when assembling from GrainSinograms. Useful if your reconstructions have meaningful relative intensities
- Optional smoothing of labels field when assembling from GrainSinograms

## `sinogram.py`
- New method: `prepare_peaks_from_4d_fast`
This is much faster than before, especially when h5 or e2dx/y files are used for spatial correction. It was slow before (for large numbers of grains) because we were reloading the same spatial distortion files from disk each time we made
 a cf_for_sino (e.g. for each grain)
- Optional structure factor correction
Added a small class into sinograms.py that handles structure factor correction. In the case where the structure is known and you have a CIF, you can normalise sinogram projections by structure factor (and LF and Polarization too while you're at it). The final normalisation is then turned off, which means this keeps meaningful intensities all the way to the final TensorMap, and is much more robust for twins.

## `fit_y0.py`
- Now much faster thanks to random downsampling of `cf_4d`, plus a more robust peak fit to get the highest standard deviation (estimate of best `y0`).
- Now Papermill-compatible
- Caches result to `ds.y0`
- Optional `doplot` parameter when running the `KDTree`

## `tomo_2_map.ipynb`
- New method in notebook for twin-friendly 4D peak assignments
Each grain takes whatever 4D peak it likes, allowing for peak overlap between grains. This is optimised to go fast.
- Building sinograms from 2D peaks (sometimes preferred) now much faster: Instantiates gves in ram once to avoid a big malloc per-grain

## other notebooks
- General `ds.y0` caching/loading
- whole_sample_mask now cached to disk


ImageD11/nbGui/S3DXRD/friedel_pair_map.ipynb
```diff
@@ -153,6 +153,19 @@
     "        ds.save()"
```
jonwright, line 153: <pre class="ql-syntax" spellcheck="false">Line #24.        d    = np.transpose((cf.dty[i1], cf.dty[i2]))            # (N,2)  RHS at y0=0
</pre><p>This doesn't look right to me. In the old code we have (dty-y0), then fit a sin + cos. In the new code with locate_pairs_affine it looks like you are fitting without the y0 correction. I don't understand the change. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/FABLE-3DXRD/ImageD11/pull/592/discussion/#comment-3665787712'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='11'/>

ImageD11/nbGui/S3DXRD/pbp_3_refinement.ipynb
```diff
@@ -79,7 +79,7 @@
     "manual_threshold = None\n",
```
jonwright, line 79: <pre class="ql-syntax" spellcheck="false">Line #1.    y0 = ds.y0
</pre><p>It probably needs a fall-back for y0 missing from dset?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/FABLE-3DXRD/ImageD11/pull/592/discussion/#comment-3665787717'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='8'/>

ImageD11/nbGui/S3DXRD/tomo_2_map.ipynb
```diff
@@ -85,10 +85,26 @@
     "# tolerance for building sinograms from assigned peaks\n",
```
jonwright, line 85: <pre class="ql-syntax" spellcheck="false">Line #21.        gs.update_recon_parameters(y0=ds.y0, shift=shift, pad=pad)
</pre><p>In this line - do we still lose the fit results by overwriting gs.recon_y0 ? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/FABLE-3DXRD/ImageD11/pull/592/discussion/#comment-3665787719'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='28'/>

jonwright, line 85: <pre class="ql-syntax" spellcheck="false">Line #1.    mask_path = os.path.join(ds.analysispath, 'whole_sample_mask.edf')
</pre><p>Isn't this a hdf file entry somewhere? Probably in the output grains file where these reconstructions will end up.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/FABLE-3DXRD/ImageD11/pull/592/discussion/#comment-3665787723'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='33'/>

jonwright, line 85: <p>OK - this is different to my lorentz + polarisation code. I assume I need to read gs.build_sinogram...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/FABLE-3DXRD/ImageD11/pull/592/discussion/#comment-3665787729'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='39'/>

jonwright, line 85: <p>didn't this happen earlier too?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/FABLE-3DXRD/ImageD11/pull/592/discussion/#comment-3665787740'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='45'/>

ImageD11/sinograms/sinogram.py
```diff
@@ -45,47 +86,24 @@ def __init__(self, grain_obj, dataset):
...
-        gv = np.transpose((cf_2d.gx, cf_2d.gy, cf_2d.gz)).astype(float)
-
-        # column to store the grain labels
+    def prepare_peaks_from_2d(self, cf_2d, grain_label, hkltol=0.25, gv=None):
+        """Prepare peaks used for sinograms from 2D peaks data.
+        Performs greedy assignment of all 2d peaks to this grain.
+        Pass a precomputed gv (shape (nrows, 3), C-contiguous float64) when
+        calling in a loop over grains to avoid rebuilding it each time."""
+        if gv is None:
+            gv = np.ascontiguousarray(
```
jonwright: OK - but this should get fixed by c2ImageD11 if it is ever ready. Just have the C code read SoA.

ImageD11/sinograms/sinogram.py
```diff
@@ -119,67 +137,104 @@ def prepare_peaks_from_4d(self, cf_4d, gord, inds, grain_label, hkltol=0.25):
 
         self.cf_for_sino = flt
 
-    def build_sinogram(self, columns=("omega",)):
+
+    def prepare_peaks_from_4d_fast(self, cf_4d, gord, inds, grain_label, spatial_corrector, hkltol=0.25):
+        """Same as self.prepare_peaks_from_4d, but uses get_colfile_from_peaks_dict_fast
+        instead of self.ds.get_colfile_from_peaks_dict, avoiding the per-grain spatial corrector reload."""
```
jonwright: I am not sure we need a function named "fast" and the old one to be kept? Just put in the speedup and forget the old code? I have no idea why the spatial is ever re-done anyway.

ImageD11/sinograms/sinogram.py
```diff
@@ -119,67 +137,104 @@ def prepare_peaks_from_4d(self, cf_4d, gord, inds, grain_label, hkltol=0.25):
...
         npks = pkmsk.sum()
         destRow = pkrow[dh[0], dh[1], dh[2], de]
         sino = np.zeros((npks, NY), "f")
         hits = np.zeros((npks, NY), "f")
         adr = (destRow * NY + iy).astype("q")
-        # Just accumulate
-        sig = self.cf_for_sino.sum_intensity.astype("f")
+    
+        # signal to accumulate, with whichever corrections are enabled
+        if do_correction:
```
jonwright: We can merge, but this smells bad to me:

I was just doing something outside:

cf.raw_intensity = cf.sum_intensity.copy() # or addcolumn
cf.sum_intensity = cf.raw_intensity * polar * lorentz * I_hkl_factor

Why is a crystallographic normalisation calculation inside a function that bins onto a sinogram?

test/papermill_test_notebooks.py
```diff
@@ -1649,14 +1649,14 @@ def test_FeAu_JADB_bb_friedel(aroot):
...
     print(papermill.__path__)
-    test_tomographic_route(destination_folder)
-    test_pbp_route(destination_folder)
-    test_FeAu_JADB_tomo(destination_folder)
-    test_FeAu_JADB_pbp(destination_folder)
-    test_FeAu_f2scan_JADB_pbp(destination_folder)
-    test_FeAu_JADB_bb(destination_folder)
-    test_FeAu_JADB_bb_grid(destination_folder)
-    test_FeAu_JADB_bb_friedel(destination_folder)
+    # test_tomographic_route(destination_folder)
```
jonwright: Doesn't this disable all the tests?

jonwright (commented):
I didn't look into the tensormap things, but the smoothing is a bit surprising. I had that in mind as an operation to perform on an existing tensormap, rather than a constructor arg. e.g.:

smoothed_tmap = raw_tmap.smooth( median_footprint = (3 , 3 ) )

Or just ndimage.filter on layers?

ImageD11/nbGui/S3DXRD/friedel_pair_map.ipynb
jadball: <p>I think the two versions are mathematically equivalent.</p><p>The new approach allows us to only do the solve once, then get (sx, sy) much faster given a y0, which helps when fitting y0.</p><p> </p><p>In both cases, for a pair of omega angles at i1, i2, we construct</p><p> </p>

<pre class="ql-syntax" spellcheck="false">R = [ -sin(w_i1)&nbsp; &nbsp;-cos(w_i1) ]
&nbsp; &nbsp; [ -sin(w_i2)&nbsp; &nbsp;-cos(w_i2) ]
</pre><p> </p><p>and the sample position:</p>

<pre class="ql-syntax" spellcheck="false">s = [ sx ]
&nbsp; &nbsp; [ sy ]
</pre><p> </p><p>We set up:</p>

<pre class="ql-syntax" spellcheck="false">d = [ dty_i1&nbsp;]
&nbsp; &nbsp; [ dty_i2 ]
</pre><p> </p><p>and a 1-vector:</p>

<pre class="ql-syntax" spellcheck="false">1 = [ 1 ]
&nbsp;&nbsp;  [ 1 ]
</pre><p> </p><p>In the original code, we form the equation:</p>

<pre class="ql-syntax" spellcheck="false">R . s = d - (y0 * 1)
</pre><p> </p><p>We solve to get:</p>

<pre class="ql-syntax" spellcheck="false">s = R^-1 . (d - (y0 * 1))
</pre><p> </p><p>We can distribute the matrix multiplication:</p>

<pre class="ql-syntax" spellcheck="false">s = R^-1 * (d - (y0 * 1))
s = (R^-1 * d) - (y0 * R^-1 * 1)
</pre><p> </p><p>We can write this as :</p>

<pre class="ql-syntax" spellcheck="false">s0 = R^-1 * d
v&nbsp; = R^-1 * 1
</pre><p> </p><p>Therefore:</p>

<pre class="ql-syntax" spellcheck="false">s = s0 - y0 * v
</pre><p>This is linear function in <code>y0</code> where <code>s0</code> is the intercept and <code>v</code> is the slope.</p><p>Both terms can be computed once and stored, and we can evaluate <code>s</code> as a function of <code>y0</code>.</p><p> </p><p>The full <code>fit_y0.ipynb</code> notebook has been tested and gets the same result as the original <code>locate_pairs</code> function.</p>

ImageD11/nbGui/S3DXRD/pbp_3_refinement.ipynb
jadball: <p>My assumption was that if you got this far (i.e. refining), then you'd have a y0 in the dataset, because it's set and saved in pbp_1</p>

ImageD11/nbGui/S3DXRD/tomo_2_map.ipynb
jadball: <p>The per-grain <code>y0</code> fit results are lost. The median value across all grains is always a better choice for refinement.</p>

jadball: <p>This has been the case for a while. I believe the idea was that if you weren't happy with the first grain's reconstruction, you could go back and modify the parameters (e.g. padding) until you were, then you'd update them for all grains before you reconstruct all grains</p><p> </p>

jadball: <p>This could indeed be in an h5 rather than an edf. The mask caching is currently only used for the tomo method notebooks, but I think we could use it for the point-by-point method too (we often use a mix of both)· The problem there is that the <code>...grains.h5</code> file might not exist if we do point-by-point first. What do you think?</p>

jadball: <p>Yup, it's applied there, and there's a small new class at the top of <code>sinogram.py</code> that handles the masking valid/invalid hkls. </p>

ImageD11/nbGui/S3DXRD/friedel_pair_map.ipynb
jadball: When we form the RHS, we are combining `d` and `1` together into `[ d | 1 ]`, so we can do a single solve on it:
```
R . s0 = d
R . v  = 1
```
therefore:
```
     R . [ s0 | v ] = [ d | 1 ]
solve(R, [ d | 1 ]) = [ s0 | v ]
```
So with one solve, we get `s0` and `v`, which we can then compose together to get `s` with any chosen `y0`.

jonwright: Is this in a comment somewhere? I missed the idea when scanning the code

ImageD11/nbGui/S3DXRD/pbp_3_refinement.ipynb
jonwright: But missing from an old pbp result? It means a new git checkout + install on the cluster can't process old datasets.

ImageD11/nbGui/S3DXRD/tomo_2_map.ipynb
jonwright: I will patch this myself later.

jonwright: We should meet with Pierre-Olivier at some point to discuss. Now there are many output h5. It belongs somewhere in recon space, so one of the grains.h5 files rather than the dataset, I think. It seems to be a per-sample property, but you could have a per-phase or per-grain mask...

jadball: I'll make it parts of the grains.h5 file for the phase we're mapping at the time

ImageD11/nbGui/S3DXRD/pbp_3_refinement.ipynb
jadball: Will add in a manual override in case ds.y0 not set

ImageD11/nbGui/S3DXRD/friedel_pair_map.ipynb
jadball: I'll clean this up and add it to the docstring

ImageD11/sinograms/sinogram.py
jadball: Will get rid of the old code. The spatial is re-done because we're re-making columnfiles from the pk2dmerge

test/papermill_test_notebooks.py
jadball: Sorry this was for my own debugging. WIll add the new tests in and re-enable.

ImageD11/sinograms/sinogram.py
jadball: Previously, normalising and sorting was in `build_sinogram`. If you like, I can fork out the normalisation to another method that works on `self.cf_for_sino` independently before we call `build_sinogram`. Perhaps we can also supply a columnfile title to `build_sinogram` to specify which column we'd like to use.

jonwright: Adding a title that defaults to 'sum_intensity' seems like the right option. It means you need hkl inside the cf_for_sino, but I think it had that anyway (or it is easy to recompute)

ImageD11/sinograms/sinogram.py
jonwright: Sounds like a bug - why aren't we using cf_2d? Are we out of memory?

Unrelated, but c2ImageD11 has ambitions to try float32 and get a factor of 2 in ram. Just need to find some money for AI tokens to finish it....

jadball:
> I didn't look into the tensormap things, but the smoothing is a bit surprising. I had that in mind as an operation to perform on an existing tensormap, rather than a constructor arg. e.g.:
> 
> smoothed_tmap = raw_tmap.smooth( median_footprint = (3 , 3 ) )
> 
> Or just ndimage.filter on layers?

I think the problem is that the TensorMap is a lossy format when compared with the original list of grainsinograms - we forget the full intensity field of the grains, and only remember the intensity field that's inside tmap.labels == grain_id for some given grain. This means that we can't recompute the intensity field (and therefore the scaled IPFs) once we decide on a new labels field. The same is true for UBI, eps_sample and anything else downstream. I guess what I'm trying to say is that the labels field should be treated as immutable after the tmap is constructed.

ImageD11/sinograms/sinogram.py
jadball: Currently the logic looks like this:

https://github.com/FABLE-3DXRD/ImageD11/blob/3775a885e3bfad04b5d6f329ccaf1e79392ff544/ImageD11/sinograms/sinogram.py#L912-L935

https://github.com/FABLE-3DXRD/ImageD11/blob/3775a885e3bfad04b5d6f329ccaf1e79392ff544/ImageD11/sinograms/sinogram.py#L90-L107

jonwright: So anywhere we have `self.ds.pk2d`, I guess it can be replaced by a `cf_2d` that has an updateGeometry already at the top of the notebook?

A detail, but perhaps we don't need to copy all of the columns.

Assuming we have a grain centre of mass position in here, it could make sense to update the g-vectors from xl,yl,zl based on the position before doing the hkl tolerance filter.

Feel free to merge something here. I guess we will come back to it later if we refactor to make a 'low memory' version that loops over grains and uses an mmap for the cf_2d (or pk2d).

jadball: I've refactored this slightly as part of the structure factor handling (adding hkls) plus a new columnfile method to copyrows only on selected columns. I think this PR is already large, we should address g-vector updating soon in a later PR

jadball:
@jonwright I've now addressed all your feedback aside from the local g-vector corrections which could be in a separate PR - thanks for the review!

jonwright (approved):

jonwright:
Thanks!

PR #593, opened 2026-07-29
Generalized omega-dependent background-stacks in `lima_segmenter.py`

AxelHenningsson:
This PR implements generalized omega-dependent background-stacks in `lima_segmenter.py` allowing for backgrounds that vary with the rotation angle of the turntable.

The story behind this upgrade is an in-situ experiment at ID11 in April 2026 in which diffuse scattering was observed from the sample environment (from two plastic support pilars in a tensile loading device). Due to the geometry of the environment this scattering occured only over a limited range in omega, and hence the background subtration must vary with scan-angle.

This PR supports dataset backgrounds as HDF5 file where the dataset `bgfile` string attribute can now be set as

```python
ds.bgfile = f"{hdf5_file_contaning_backgrounds}::{hdf5_data_path_to_bg_frames}"
```

where `ds` is a `ImageD11.sinograms.dataset` object.

NOTE: This PR is backwards compatible, i.e .edf backgrounds still work as before.

Cheers
Axel

### DIFF
--- a/ImageD11/nbGui/segmenter_gui.py
+++ b/ImageD11/nbGui/segmenter_gui.py
@@ -1,38 +1,70 @@
-import numpy as np
-import matplotlib.pyplot as plt
-from ipywidgets import interact, interactive, widgets, fixed, Layout
-from IPython.display import display
 import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+from IPython.display import display
+from ipywidgets import widgets
+
 import ImageD11.sinograms.lima_segmenter
 import ImageD11.sparseframe
-from ImageD11.nbGui.segmenter_gui_utils import * 
+from ImageD11.nbGui.segmenter_gui_utils import *
+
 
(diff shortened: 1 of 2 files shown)


jonwright:
Thanks!

Is there any way to see a diff with the formatting changes removed? As it reads, there is a bunch of stuff I don't understand mixed with a lot of noise. Perhaps something like 'winmerge' where you refuse all the changes that are not fixing the background

Some examples: 
```
ref = frms[self.idx] != frms[i+self.idx]

10**howmany != int(round(2**howmany))
```
Where you have "bg_is_stack": this is not a SegmenterOption. You give the bgfile and the file tells us what it is (single frame, multi-frame). No need to change the API.

Was this a re-write by an LLM? Can you ask it to make a minimal diff that only modifies for the background code, touching nothing else, and refactor that to go via the file?

jonwright:
Perfect - thanks!

PR #613, opened 2026-08-28
pbp_refine correctness fixes + voxel_mask-based speedups for pbp indexing/refinement

jadball:
Sorry for the huge PR. This is many small fixes and some big ones.
Disclaimer: **Quite a bit of the heavy lifting to integrate voxel_mask was done with Claude**, although I have carefully fought back against stupid decisions, checked it and documented. I've ran through the FeAu pipeline to vaildate, and also tried on some larger datasets ( some round-robin data, and thanks to @adcret ). Things should work. I have users next week I will try this on too.

Big thanks to @AxelHenningsson  for `voxel_mask.py`, which most of the speedups below are built on. Sorry we didn't pick this up sooner!

### Breaking changes

- Beam weight was wrong in `pbp_refine`. The distance weighting used in both fits, `w = r / (y + dy + r)`, had `r` hard-coded to `1.0`, intended to mean "1 micron" no matter what units your `dty` motor actually reports in. On a dataset scanned in mm (TDXRD) this makes the weighting go flat, so every peak along a ray counts equally regardless of how far the beam actually passed from the voxel centre. `r` is now `beam_size / 3`, and `beam_size` is a real parameter on `PBPRefine` (defaults to `ystep`). To reproduce old results exactly on a 3 micron beam experiment, pass `beam_size=3`.
- `hkl_tol_origins` was wrong in `pbp_refine`. Origin finding was comparing the squared hkl residual against an unsquared tolerance, so a nominal `hkl_tol_origins=0.05` was actually being applied as `sqrt(0.05) ~= 0.224`. It is now squared consistently with `hkl_tol_refine`, so `0.05` really means `0.05`. Origins computed with older code used the looser tolerance.
- `sel` vs `isel` mix-up in pbp indexing and refinement. When building the working peak table (`icolf`), both `point_by_point.py` and `pbp_refine.py` were copying rows using `sel` (the ring / intensity-fraction mask) instead of `isel` (`sel` further filtered by `etacut`). Peaks that should have been excluded by the eta cut were silently making it into indexing and refinement. Both now consistently use `isel`.
- NaN voxels from an unguarded matrix inversion in `refine_map`. `ubi_and_ucell_to_u` called `np.linalg.inv` on the metric tensor. For a poorly-conditioned UBI this raised a `LinAlgError` inside a multi-threaded `prange`; that exception isn't reliably propagated, so the affected thread just stopped, leaving every voxel it still owned as NaN. Since voxels are chunked across workers, this showed up as whole strips of NaN voxels rather than isolated speckle, worth a look if you've seen anything like that in older maps. Replaced with a guarded closed-form version that skips the strain fit for that voxel instead of raising.
- `sbatchlocal` off-by-one in `lima_segmenter.py`. SLURM `--array=start-end` is inclusive of `end`, but the range used to build local commands went to `end` (exclusive), so the last array task was silently skipped, and a single-job array (`--array=0-0`) would run nothing at all. Fixed to `range(start, end + 1)`.
- `sino_shift_and_pad` was half a pixel out for odd `ny`. It took the middle of the sinogram to be `ny/2`, but the reconstruction origin sits on integer index `ny//2`. Those agree for even `ny` and differ by 0.5 for odd, so every reconstruction from a scan with an odd number of `dty` bins landed half a pixel from where `sample_to_recon` and `recon_to_sample` said it should. This hit iradon and ASTRA equally, which is why it never showed up as a disagreement between the two. Now uses `ny//2`. Worth noting the Si cube test dataset has `ny=41`, so it was affected.
- Reconstructions are now always odd-sized. `sino_shift_and_pad` bumps `pad` by one when `ny + pad` would come out even, so the rotation axis at index `N//2` is always the geometric middle pixel of the output. This is what lets `recon_to_step`, `recon_bins` and ASTRA agree without anyone applying a half-pixel correction: ASTRA centres its volume on `(N-1)/2`, which only equals `N//2` for odd `N`. Reconstruction sizes therefore change by one pixel for scans that previously produced an even size, and downstream code that hard-codes a shape will need updating.
- ASTRA reconstructions were offset from the iradon ones. `run_astra` handed ASTRA the raw sinogram, where `run_iradon` pads it internally out to the diagonal of the output. Two things followed: the ASTRA detector was only `ny` wide so it did not span the reconstruction volume, and grains at a radius approaching `ny/2` fell outside it; and `shift_sinogram` moved data within a fixed-size array, so a large `shift` slid real signal off the end and discarded it (55% of the sinogram intensity in the half-acquisition test case). `run_astra` now pads the sinogram the way iradon does and keeps the padded detector length odd so its centre lands on an integer index. Measured against `sample_to_recon` on a half acquisition the grain went from about 1 px off to within 0.03 px, matching iradon. Half acquisitions and grains far from the rotation axis were worst affected; a well-centred full scan was already fine.
- Large `y0` errors were mis-reconstructed by `run_iradon` and `run_mlem`. `projection_shifts` was zero-padded, so the padded rows were declared unshifted while the real detector rows carried `shift`. That makes the sample coordinate array non-monotonic, and `np.interp` returns nonsense on non-monotonic input without warning. Negligible for `|shift| <= 1`, but about 40% of peak intensity by `|shift| = 15`. Now padded with `mode='edge'` so the shift extends into the pad.
- Masks were silently dropped for `FBP_CUDA` and `BP_CUDA`. Both inherit from `CCudaReconstructionAlgorithm2D`, whose initialisation consumes `ReconstructionMaskId` without any algorithm ever applying it, so unlike CPU `FBP` (which at least warns about the unused option) they accepted a mask, ran cleanly, and reconstructed unmasked. Both now go through the same manual masking as `FBP` and `EM_CUDA`.
- `run_astra` no longer clamps reconstructions to `[0, 1]` by default. `MinConstraint` and `MaxConstraint` were hard-coded, which silently saturated the output whenever the sinogram was not normalised to a peak of 1 (`build_sinogram(normalise=False)`, or after `correct_ring_current`). Non-negativity is kept, but the upper clamp is now opt-in via the new `max_constraint` argument. Pass `max_constraint=1` to reproduce old behaviour.
- Half-scan masking rewritten. `apply_halfmask_to_sino` previously assumed the rotation axis sat at `len(sino)//2 - 1`; it now takes the axis row explicitly (defaulting to `ny//2`, where `correct_bins_for_half_scan` puts it), weights the axis row 0.5 because it is measured at every omega, and decides which side was actually scanned from `ds.ybin_real_mask` rather than guessing. Half-scan reconstructions will shift by one row relative to before.
- `correct_bins_for_half_scan` now updates `ds.ymin` and `ds.ymax`. They were only ever set in `guessbins`, so once half-scan padding moved `ybincens[0]`, everything reading `ds.ymin` (`recon_bins`, `guess_y0`, `dty_to_dtyi`) was out by the width of the low-side padding.

### Big changes

- `pbp_refine` moved into its own file, `ImageD11/sinograms/pbp_refine.py`, separate from `point_by_point.py`. Import path changes: `from ImageD11.sinograms.pbp_refine import PBPRefine` instead of importing it from `point_by_point`.
- Both pbp indexing and pbp refinement are substantially faster, because they no longer scan every peak per voxel to work out what is relevant. Instead,  we build a partition of the peaks over an (omega, dty) grid using `voxel_mask`: peaks are binned and sorted so that, for any voxel, only the strip of cells its sinusoid can actually reach needs visiting, instead of the whole map. The exact per-peak geometric test is still applied on top of that pre-filter, so results are unchanged, only the search is cheaper. For pbp indexing specifically, this partition is then written to disk (`index_filename`) and memory-mapped read-only by each worker process, the same way `icolf` already is via `mmap_h5colf`. Since the partition is identical in every worker, one mapping through the page cache replaces what would otherwise be a separate in-memory copy per process, and this is also what lets the same indexing job be split across a SLURM array. g-vector computation is also faster now, working directly from peaks' lab-frame vectors rather than recomputing them from scratch each time. Measured on a 3x10^6-peak dataset (Ryzen 7840U): `pbp_refine`'s refinement step went from 2m51s down to about 11s with a warm cache, and origin-finding to about 4.8s. This was before the gve computation speedups, so real improvements may be even faster. Both kernels were checked against the previous implementation: `refine_map` gives identical `npks`/`nuniq` and agrees on UBI and strain to 1e-14 relative; `compute_origins` finds the same set of origins, agreeing to 4e-16.
- ASTRA reconstruction can now run locally, not just via SLURM. The `tomo_2_map` / `tomo_2_map_minor_phase` notebooks gained a `run_astra_on_cluster` toggle: off, they run ASTRA in the notebook process directly, with a check that you actually have CUDA if you asked for a CUDA method.
- ASTRA gained `FBP` support, plus plain (non-CUDA) `BP` and `SIRT`, which run on CPU rather than needing a GPU. Only CPU `FBP` needs the sinogram resampled, because it is the one method that rejects the vec geometry `astra.functions.geom_postalignment` produces; everything else, `FBP_CUDA` included, still shifts the geometry and never touches the intensity data. That resampling uses linear rather than cubic interpolation, since cubic overshoots and can hand ASTRA negative intensities. `run_astra_recon.py` and `prepare_astra_bash` now take the ASTRA method as an argument instead of it being hard-coded to `EM_CUDA`.
- `apply_halfmask` and `mask_central_zingers` moved up into `GrainSinogram.recon`. They are properties of the sinogram and of the finished reconstruction rather than of the backend, and previously only `run_iradon` and `run_mlem` accepted them, so passing either alongside `method="astra"` raised `TypeError`. They now work with every backend.

### Notebook changes

- New `dsfix.ipynb`: lets you repoint a previously-segmented dataset at a new `analysisroot` / spatial-correction files, for when you want to move processing to a different location than where segmentation ran.
- `pbp_1_indexing.ipynb`, `pbp_3_refinement.ipynb` and `tomo_3_refinement.ipynb` gained a `drop_input_peaks` option: for large datasets, you can now free `cf_2d` from memory once the peaks needed for indexing/refinement have been selected out of it.
- `0_segment_and_label.ipynb`: local segmentation option added, as an alternative to submitting to the cluster.
- `tomo_2_map.ipynb`: guesses `y0` if none is supplied, rather than proceeding with a bad default.
- `nb_utils.py`: new `plot_grain_recons`, an interactive slider viewer for stepping through grain reconstructions and sinograms side by side.

### Small changes

- `unitcell.py`: `gethkls_xfab` used to return hkls in whatever order xfab's `genhkl_all` happened to produce, which wasn't consistent between runs. It's now lexsorted, so the ring order is deterministic.
- `unitcell.py`: `makerings` is cached if called again with the same `limit`/`tol` as last time, instead of rebuilding `ringds`/`ringhkls` from scratch every call. Matters because pbp indexing calls it repeatedly via `assigntorings`.
- `unitcell.py`: new `getcoses`, caching the distinct cosines between two rings' hkls per ring pair instead of rebuilding the list on every call; `indexing.py`'s `indexer.find` now calls this instead of duplicating the same loop inline.
- `unitcell.py`: `orient` now uses a closed-form 3x3 matrix inverse (`_inv3`) instead of `np.linalg.inv`.
- `indexing.py`: `refine`'s normal equations are now built with matrix multiplies instead of per-peak `np.outer` calls in a loop. Numerically equivalent (results differ by ~1 ULP from a different BLAS accumulation order), just faster.
- `indexing.py`: `indexer.reset()`'s pristine-state snapshot no longer deep-copies the shared `unitcell` object, which used to drag its `ringhkls`/`anglehkl_cache` along for ~2000 nested copies per indexer construction; `reset()` also now updates `__dict__` from that snapshot instead of replacing it and re-copying, so it deep-copies once per call instead of twice.
- `peakselect.py`: `mask_rings_by_ifrac` gained optional `verbose` and `return_npks_hmax` arguments (default off, backwards compatible).
- `properties.py`: peak tables now also store the frame number (`frm`), used by the new omega-binning logic in `voxel_mask`.
- `run_pbp_recon_chunk.py`: worker config now carries an `index_filename`, matching the new `voxel_mask` index caching.
- `sinogram.py`: new `_astra_geoms` builds the ASTRA volume and projection geometry from the unpadded sinogram height, so the reconstruction (and any mask) keeps the shape it had before while the detector alone is widened.
- `sinogram.py`: `"EM"` dropped from `run_astra`'s allowed methods. ASTRA only registers `EM_CUDA`, so a CPU `EM` request could never do anything but fail with "Unknown algorithm type".
- `sinogram.py`: `astra.projector.delete(proj_data_id)` corrected to `astra.data2d.delete`, since that id comes from `data2d.create`; and `if astra_method != ["FBP"]` corrected to compare against a string rather than a list.
- `geometry.py`: the module docstring now says where the origin actually is. Integer indices label pixel centres in step space, reconstruction space and `dtyi` alike; the rotation axis lands on index `N//2`, which is the middle pixel for odd `N` and half a pixel past the geometric centre for even `N`; and there's a new section describing what `shift` means and how `sino_shift_and_pad` defines it.
- `4_visualise.ipynb`: `cm.get_cmap` replaced with `plt.get_cmap` (removes a matplotlib deprecation warning).
- `test_point_by_point.py` retired; replaced by `test_pbp_refine.py` and `test_voxel_mask.py`. New `test_sino_recon.py`, built out of `sandbox/scanning_geometry.ipynb`: full (180 degrees across the whole sample) and half (360 degrees across half) acquisitions crossed with `y0` errors, reconstructed through both iradon and ASTRA and checked against `sample_to_recon` to a third of a pixel, with mask handling checked per method and the odd-size invariant of `sino_shift_and_pad` pinned down directly. Even output sizes are still covered by forcing extra padding, so the half-pixel path stays exercised. CUDA methods skip with a warning when no GPU is available, and the ASTRA tests skip entirely if astra isn't installed. `papermill_test_notebooks.py` updated for the new notebooks and options above.


jadball:
Please hold off on merging for now. Chasing some half-voxel shift bugs in astra vs iradon recons.

jadball:
I'm now happy to merge

ImageD11/unitcell.py
```diff
@@ -532,6 +552,12 @@ def makerings(self, limit, tol=0.001):
         The tolerance is the difference in d* to decide
         if two peaks overlap
         """
+        # Cache if limit and tol are the same
+        # this is called each time by assigntorings
+        # slow if we are doing pbp indexing
+        if (getattr(self, "_rings_for", None) == (limit, tol)
```
jonwright, line 558: todo: set rings_for to limit, tol

ImageD11/indexing.py
```diff
@@ -655,24 +669,25 @@ def find(self):
...
+        #     if (
+        #         abs(a - 1.0) < 1e-5 or abs(a + 1.0) < 1e-5
+        #     ):  # Throw out 180 degree angles
+        #         continue
+        #     if len(coses) == 0:
+        #         coses.append(a)
+        #         continue
+        #     if abs(coses[-1] - a) > 1e-5:
+        #         coses.append(a)
+        coses = self.unitcell.getcoses(self.ring_1, self.ring_2)
```
jonwright, line 691: requires cos_tol ?? Why was this 1e-5

ImageD11/sinograms/voxel_mask.py
```diff
@@ -1,13 +1,406 @@
...
+# and also going between a local machine and a cluster machine.
+
+_CACHE_VERSION = 7
+
+
+## Save and load caches
+_PEAK_H5GROUP = "PBPPeakIndex"
+
+
+def default_index_filename(manager):
```
jonwright, line 59: Isn't this icolf in dataset?

ImageD11/sinograms/point_by_point.py
```diff
@@ -33,18 +31,18 @@
...
 from ImageD11.columnfile import columnfile
+from ImageD11.peakselect import mask_rings_by_ifrac
 from ImageD11.sinograms import geometry
-from ImageD11.sinograms.roi_iradon import run_iradon
-from ImageD11.sinograms.tensor_map import unitcell_to_b
-from ImageD11.sinograms.sinogram import save_array
-
+from ImageD11.sinograms.voxel_mask import (
+    VoxelSinoMasker, fill_voxel_idx, max_candidates as vm_max_candidates,
+    default_index_filename,
```
jonwright, line 42: move?

ImageD11/sinograms/point_by_point.py
```diff
@@ -1713,83 +274,165 @@ def get_local_gv(si, sj, ystep, omega, sinomega, cosomega, xl, yl, zl):
...
+            out[t] = idx_buf[t]
+        return n
+ 
+    m = 0
+    inv_ystep = 1.0 / ystep
+    for t in range(n):
+        q = idx_buf[t]
+        dty_calc = y0 - sx * sinomega[q] - sy * cosomega[q]
+        # round() here matches numpy's round-half-to-even, as dty_to_dtyi uses
+        if int(round((dty_calc - ymin) * inv_ystep)) == dtyi[q]:
```
jonwright, line 341: Isn't it return `idx_buf[ ydist_buf <= ystep/2 ]` ?

ImageD11/sinograms/point_by_point.py
```diff
@@ -1713,83 +274,165 @@ def get_local_gv(si, sj, ystep, omega, sinomega, cosomega, xl, yl, zl):
...
+ 
+    n = fill_voxel_idx(sx, sy, y0, ystep, ymin,
+                       omega_partitions, dty_partitions, dty,
+                       sinomega, cosomega, sinomega_bins, cosomega_bins,
+                       idx_buf, ydist_buf)
+    if n < 0:  # buffer too small, catch and return -1 to the caller
+        return -1
+ 
+    if relax:  # match voxel_mask predicate, so just return the peaks that fill_voxel_idx() found
+        for t in range(n):
```
jonwright, line 331: is a copy needed? Can we avoid numba here

ImageD11/sinograms/point_by_point.py
```diff
@@ -2000,26 +715,44 @@ def setpeaks(self, colf, icolf_filename=None):
         print(
             "Using for indexing:",
             self.icolf.nrows,
-            "npks, minpks, forgen",
-            self.npks,
+            "npks, minpks, forgen, foridx",
+            npks,
```
jonwright, line 719: was self.npks before? Do we need it?

ImageD11/sinograms/point_by_point.py
```diff
@@ -2098,6 +832,8 @@ def submit_slurm_chunks(self, grains_prefix, id11_code_path, gridstep=1, n_chunk
 #SBATCH --cpus-per-task={cpus_per_chunk}
 #SBATCH --array=0-{array_end}
 #SBATCH --mem={mem_G}G
+export NUMBA_CACHE_DIR=/tmp/numba_${{SLURM_JOB_ID}}_${{SLURM_ARRAY_TASK_ID}}
```
jonwright, line 835: Should be generic to all jobs?

jonwright: And deletion?

jonwright (commented):
see the rest later

jonwright:
Merging. Will put some edits on top soon.

PR #617, opened 2026-09-04
Store pbp peak index as a group inside the icolf file

jonwright:
TEST THIS BEFORE MERGING: AI WRITTEN CODE

The voxel_mask peak index is a lookup/partition table for the icolf it accompanies, so store it as the PBPPeakIndex group in the same HDF5 file instead of a sibling _pbpindex.h5 sidecar file. The peaks group is left untouched; mmap_h5colf and colfile_from_hdf find columns by ImageD11_type, so the extra group does not confuse them.

Remove default_index_filename and the PBP.index_filename attribute / point_by_point index_filename parameter; the partition is now always read from the icolf file in the initializer, warm-up and SLURM chunk worker.


jadball:
seems good to me. I have something similar but not pushed. happy to merge

PR #622, opened 2026-09-06
sparseframe: add optional smoothing to sparse_localmax

### LINKED ISSUE #159: SparseFrame vs SparseScan and lmlabel
The smoothing code only applies on SparseScan and not sparse_frame - clean this up.

jonwright:
Fixes #159. SparseScan.lmlabel already smooths via cImageD11.sparse_smooth before localmax labelling, but the frame-level equivalent sparse_localmax() had no way to do the same. Adds a smooth=False kwarg (default keeps existing behaviour/output unchanged) that reuses the existing sparse_smooth helper, same as SparseScan.lmlabel already does.

### DIFF
--- a/ImageD11/sparseframe.py
+++ b/ImageD11/sparseframe.py
@@ -633,12 +633,19 @@ def sparse_connected_pixels(
     return nlabel
 
 
-def sparse_localmax(frame, label_name="localmax", data_name="intensity"):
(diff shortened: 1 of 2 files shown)


jonwright:
I think there is zero test coverage for this. Lets add a test please

PR #626, opened 2026-09-06
Issue625

jonwright:
Cleans up a bunch of shape= things for numpy 2.5. 

REQUIRES REVIEW - LLM code

### DIFF
--- a/ImageD11/depreciated/peaklabel.py
+++ b/ImageD11/depreciated/peaklabel.py
@@ -158,9 +158,9 @@ def build_overlap_matrix(ds, hname):
     npk = ds.nlm.sum()
     ds.firstpk = np.zeros( ds.nlm.shape, int )
     s = ds.firstpk.shape
-    ds.firstpk.shape=-1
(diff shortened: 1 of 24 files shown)


jonwright:
I am still not happy with the nputils code. If we are running a free threaded python then I think we should require the reshape code path. So if someone installs numpy < 2.1 (or is it <2.5) on free threaded, then their env is not great for free threaded and everything will be disappointing. So we should check if we are on a free threaded build, and if we are, shall we insist that the numpy is compatible. If it is not, we should issue a logging.warning that they have a broken environment. The question is whether to raise an Exception, and refuse to run (e.g. force them to have latest numpy) or whether we allow people to use the code in broken envs.... I think this depends on how numpy 2.0 and 2.1 behave with 3.14t. The shape setting is a hypothetical bug which may or may not be a problem, depending on race conditions. It does not always come up. Whichever way we go, I don't think it is good for me to have ImageD11 running on 3.15t with a numpy that allows `ar.shape=` and not be aware of that happening. What can you propose?

jonwright:
This should fix the numpy<2 error in polefigures.

PR #632, opened 2026-09-15
cf_4d Friedel pair indexing notebook

jadball:
Finally adding the Friedel pair indexing notebook (similar to that I used for the Ni superalloy paper). This is a new route for indexing that you could pick instead of `tomo_1_index`. It has support for automatic sigma 3 twin generation too, if you want.

This often performs very well for annealing twins, see this 316 stainless example:

image


jadball:
@jonwright happy to merge?

jonwright:
Maybe a future this: @jbjacob94 : can we merge some of the python functions in here into friedel_pairs.py?

jonwright:
And I am realising sonnet is quite far behind the frontier. We need to get some better testing set up...

jbjacob94:
@jadball @jonwright There is already a function "locate_eta_pairs" which is doing something similar to locate_pairs_affine in friedel_pairs.py. I am checking now whether it is returning the same coordinates. I did probably mess around a little bit with the lab / sample reference frames. 

For finding the pairs, find_pair_minus_eta in the notebook is probably good enough for simple cases but may not scale up very well for very large peak files (because of kd-trees). I would recommend using the functions in friedel_pair.py for this, where you can split the columnfile by eta bins and match by pairs of opposite bins. 

edit: I checked the pair relocation functions. They both compute the same thing but locate_eta_pairs in the friedel pair module takes dty - y0 as rhs, so it returns directly sx and sy, while locate_pairs_affine returns a s0 independent on y0. In the end doing sx,sy = s0 - y0*v falls back to the same result.

jonwright:
OK, lets merge for now and make a note to cleanup later

Issue #602, opened 2026-07-31
Remove shm from sinograms.properties


jonwright:
... or ensure the oom killer does not leave a mess.

Maybe easiest fix is to use threaded and drop multiprocessing (h5io??)

Issue #479, opened 2025-04-15
merging omega in sparse pixels  mod 360


jonwright:
A bug / glitch for mod360 case: @jadball has a fix for omega split over 360 degrees (merge sin/cos omega and take arctan2 on the result).

Issue #469, opened 2025-03-26
Dataset peaks table memory usage


jonwright:
peaks_table stays in memory inside dataset

An example here:
/data/visitor/ma5608/id11/20250212/Wenxi/PROCESSED_Ti/NS_Ti_deformed/NS_Ti_deformed_Z600_3/JWtest-lowmem-peakstable-selector.ipynb

Deleting it frees up a lot of ram.

Perhaps there is some way to load the peaks without keeping pk_props, etc, in memory?

Issue #265, opened 2024-03-19
Get rid of the shared memory arrays : add a colfile mmap option


jonwright:
In sinograms/properties.py and sinograms/point_by_point.py the code works via shared memory.
This does not scale beyond one node and it fails on python2.7.

For global read-only memory we could use mmap with numpy on a non-compressed hdf5 file (https://gist.github.com/maartenbreddels/09e1da79577151e5f7fec660c209f06e):

```
assert dset.chunks is None and dset.compression is None and dset.is_virtual is None and dset.is_external is None and etc.
file = open(path, "rb")
fileno = file.fileno()
mapping = mmap.mmap(fileno, 0, access=mmap.ACCESS_READ)
np.frombuffer(mapping, dtype=dset.dtype, count=dset.size, offset=dset.id.get_offset()).reshape(dset.shape)
```

This may be useful for reducing some out-of-memory problems. 

Another upgrade path could be looking into dask.dataframe for distributed processing.

jonwright:
Note : multiprocessing + shared memory seems to be buggy. The remove from tracker monkeypatch does not work. Abandon it.
```
Exception ignored in: <Finalize object, dead>
Traceback (most recent call last):
  File "/cvmfs/hpc.esrf.fr/software/packages/linux/x86_64/jupyter-slurm/2023.10.7/envs/jupyter-slurm/lib/python3.11/multiprocessing/util.py", line 224, in __call__
    res = self._callback(*self._args, **self._kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/cvmfs/hpc.esrf.fr/software/packages/linux/x86_64/jupyter-slurm/2023.10.7/envs/jupyter-slurm/lib/python3.11/multiprocessing/synchronize.py", line 87, in _cleanup
    sem_unlink(name)
FileNotFoundError: [Errno 2] No such file or directory
```

jonwright:
There is still the evil monkeypatch in properties.py

jonwright:
Within #389 the monkeypatch is suppressed by:
`if os.environ.get('IMAGED11_PATCH_MP', None) != 'MONKEYPATCH':`

This is not ideal, but it appears to run with or without the patch in cases where the code works. This could be about cleanup during OOM problems.

Cleaning all this up would be a natural case for also developing #252

jonwright:
[...]

PR #633, opened 2026-09-15
Unify frelon and eiger peak merging via sparse pixel files

jadball:
Claude vomit. Careful review needed, although I have ran the frelon segmentation notebook on box-beam 3DXRD data and it seemed to work!

Frelon segmentation now writes the same sparse pixel file as the eiger route (assemble_label layout), including the per-frame peak labels, so 2D peaks from both detectors are labelled and merged by sinograms.properties instead of the slow KD-tree do3dmerge/do4dmerge.

- frelon_peaksearch: worker.peaksearch_sparse and segment_dataset_to_sparse (streams pixels + labels per scan; monitor used for thresholding, raw intensities stored, normalisation applied later via ds.monitor). segment_dataset is deprecated.
- sinograms.properties:
  * algorithm='stored' uses labels saved by the segmenter
  * option max_centroid_dist: overlapping 2D peaks are only merged if their centroids are within this many pixels (None = any overlap, unchanged default). 1.6 reproduces the old frelon merge.
  * pixel weights are float (was int64, truncating float pixels); integer data give identical results
  * merge options recorded as pks2d attributes
- sparseframe.SparseScan reads stored labels + nlabel when requested
- assemble_label: write_scan_header split out and shared
- notebooks: frelon S3DXRD/TDXRD use the sparse route with max_centroid_dist = 1.6; eiger notebook exposes the option (None)
- test/test_sparse_merge.py: gate unit tests, and old vs new frelon 3D/4D merge comparison (with monitor) on a synthetic bliss dataset


ImageD11/sinograms/assemble_label.py
```diff
@@ -162,41 +205,10 @@ def harvest_masterfile(
...
-                        ds[()] = data
-                gip = g.require_group("instrument/positioners")
-                for m in headermotors:  # fixed : scalar
-                    if "instrument/positioners/%s" % (m) in gin:
-                        data = gin["instrument/positioners"][m][()]
-                        ds = gip.require_dataset(m, shape=data.shape, dtype=data.dtype)
-                        ds[()] = data
-                try:
-                    frms = gin["measurement"][dset.detector]
-                except Exception as e:
```
jonwright, line 190: I don't see these in the refactor?

ImageD11/sinograms/properties.py
```diff
@@ -215,7 +241,9 @@ def props(scan, i, algorithm="lmlabel", wtmax=None):
         e = s + scan.nlabels[j]
         # [1:] means skip the background labels == 0 output
         r[0, s:e] = np.bincount(f0.pixels["labels"])[1:]
-        wt = f0.pixels["intensity"].astype(np.int64)
+        # float weights: segmenters may store background subtracted floats.
+        # (was int64, which truncates each pixel. Integer data are unchanged.)
+        wt = f0.pixels["intensity"].astype(np.float64)
```
jonwright, line 243: This needs a TODO: the type should follow the input (int vs float)

ImageD11/sinograms/properties.py
```diff
@@ -429,6 +457,9 @@ def save(self, h5name, group="pks2d", rc=False):
         opts = {}  # No compression is faster
         with h5py.File(h5name, "a") as hout:
             grp = hout.require_group(group)
+            # record how the peaks were labelled and merged
+            for k, v in getattr(self, "merge_options", {}).items():
```
jonwright: pks_table does not have a merge_options member in init. What is in here?

ImageD11/sinograms/properties.py
```diff
@@ -511,30 +542,40 @@ def load(cls, h5name, h5group="pks2d"):
         obj.npk = npk  # this is ugly. Sending as arg causes allocate.
         return obj
 
-    def find_uniq(self, outputfile=None, use_scipy=False):
-        """find the unique labels from the rc array"""
+    def find_uniq(self, outputfile=None, use_scipy=False, keep=None):
+        """find the unique labels from the rc array
+
+        keep = optional boolean mask over the columns of self.rc.
```
jonwright: I do not understand this. What is the intention here?

ImageD11/sinograms/properties.py
```diff
@@ -511,30 +542,40 @@ def load(cls, h5name, h5group="pks2d"):
...
+               Only pairs with keep == True are used to merge peaks.
+               (e.g. from centroid_gate). self.rc itself is not modified.
```
jonwright: Centroid == middle of a peak.
Gate == Claude's way to say "if statement" or "mask" I think.

ImageD11/sinograms/properties.py
```diff
@@ -511,30 +542,40 @@ def load(cls, h5name, h5group="pks2d"):
...
+        """
         t = tictoc()
         n = self.ipk[-1]
+        rc = self.rc
+        if keep is not None:
+            keep = np.asarray(keep, dtype=bool)
+            assert keep.shape == (rc.shape[1],), "keep must match rc.shape[1]"
+            rc = rc[:, keep]
```
jonwright: This filter belongs outside this function call I think.

ImageD11/sinograms/properties.py
```diff
@@ -713,6 +754,46 @@ def get_clean_labels(labels):
     return n
 
 
+@numba.njit(parallel=True)
+def centroid_gate(rc, pk_props, maxdist):
```
jonwright: rename please.

jonwright: You can filter on the number of overlapping sparse pixels too (normalise to the number of peaks in each peak)

ImageD11/sinograms/properties.py
```diff
@@ -713,6 +754,46 @@ def get_clean_labels(labels):
...
+    for k in numba.prange(npairs):
+        i = rc[0, k]
+        j = rc[1, k]
+        dr = pk_props[2, i] / pk_props[1, i] - pk_props[2, j] / pk_props[1, j]
+        dc = pk_props[3, i] / pk_props[1, i] - pk_props[3, j] / pk_props[1, j]
+        keep[k] = (dr * dr + dc * dc) <= d2max
+    return keep
+
+
+def merge_mask(pkst, max_centroid_dist=None, verbose=1):
```
jonwright: delete the function

ImageD11/sinograms/properties.py
```diff
@@ -192,18 +192,44 @@ def pairscans(s1, s2, omegatol=0.051):
     return pairs
 
 
+ALGORITHMS = ("lmlabel", "cplabel", "stored")
+
+
+def open_scan(hname, scan, algorithm="lmlabel"):
```
jonwright: Delete the function and use SparseScan directly

ImageD11/sinograms/properties.py
```diff
@@ -732,20 +813,24 @@ def find_ND_labels(i, j, npks, verbose=1):
...
+    max_centroid_dist = None (merge any overlapping peaks)
+                        or only merge overlapping peaks with centroids within this many pixels
 
     returns a pks_table.
         You might want to call one of "save" or "pk2d" or "pk2dmerge" on the result
 
     This is probably not threadsafe
     """
-    sps = ImageD11.sparseframe.SparseScan(sparsefilename, ds.scans[row])
+    sps = open_scan(sparsefilename, ds.scans[row], algorithm)
```
jonwright: see above. If you need a change in SparseScan propose it there. I think you require a "stored" column for "stored" labels?

ImageD11/sinograms/properties.py
```diff
@@ -777,7 +862,8 @@ def pks_table_from_scan(sparsefilename, ds, row, algorithm='lmlabel', wtmax=None
         rc[1, s:e] = pkid[frame2] + ijn[:, 1] - 1  # row
         rc[2, s:e] = ijn[:, 2]  # num pixels
         s = e
-    uni = pkst.find_uniq()
+    assert s == rc.shape[1]
+    uni = pkst.find_uniq(keep=merge_mask(pkst, max_centroid_dist))
```
jonwright: As I read the code - everybody get max_centroid_dist whether they like it or not?

The sparse graph filter (or not) should be an operation that is distinct from find_uniq (and where I can opt out)

ImageD11/sinograms/properties.py
```diff
@@ -818,7 +904,7 @@ def process(qin, qshm, qout, hname, dsfilename, options):
     prev = None  # suppress flake8 idiocy
     # This is the 1D scan within the same row
     for i in range(start, end + 1):
-        scan = ImageD11.sparseframe.SparseScan(hname, scans[i])
+        scan = open_scan(hname, scans[i], options["algorithm"])
```
jonwright: see above.

ImageD11/sinograms/properties.py
```diff
@@ -929,10 +1015,11 @@ def goforit(ds, sparsename, options):
 
 
 default_options = {
-    "algorithm": "lmlabel",
+    "algorithm": "lmlabel",  # | cplabel | stored
     "wtmax": None,  # value to replace saturated pixels
     "save_overlaps": False,
     "nproc": None,
+    "max_centroid_dist": None,  # None == merge any overlap, else pixels
```
jonwright, line 1028: this is the opt out, right?

ImageD11/sinograms/properties.py
```diff
@@ -943,20 +1030,27 @@ def main(dsfilename, sparsefile=None, pksfile=None, options={}):
...
-            default_options[k] = options[k]
+        if k in opts:
+            opts[k] = options[k]
         else:
             raise Exception("I do not understand %s in options" % (str(k)))
-    options = default_options
+    options = opts
+    if options["algorithm"] not in ALGORITHMS:
+        raise ValueError("algorithm must be one of %s" % (str(ALGORITHMS)))
+    if options["max_centroid_dist"] is not None and options["max_centroid_dist"] <= 0:
```
jonwright, line 1060: There are other filters. Louise was keen on defining these too.
The rc array is a sparse distance matrix, you can invent various other distance functions.

These are two orthogonal changes for me. One is the stored labels, the other is the distance selections.

ImageD11/sinograms/properties.py
```diff
@@ -989,13 +1083,19 @@ def main(dsfilename, sparsefile=None, pksfile=None, options={}):
...
+            cc = rmem.find_uniq(keep=merge_mask(rmem, options["max_centroid_dist"]))
             t("%s connected components" % (str(cc[0])))
         else:
             # single scan. Skips a lot of hassle.
             rmem = pks_table_from_scan(sparsefile, ds, 0,
                                        algorithm=options['algorithm'],
-                                       wtmax=options['wtmax'])
+                                       wtmax=options['wtmax'],
+                                       max_centroid_dist=options['max_centroid_dist'])
+        rmem.merge_options = {
```
jonwright: Belongs in init?

jonwright (commented):
Mostly comments aimed at Claude.
I was hoping we can refactor all of this soon. 
Putting the matrix cut inside the labelling is not good for me. 

I want to be able to build and label sub-matrices in the future, and merge them later, so it doesn't work to expect all of rc to always be in memory and amenable to a selection operation. It might also be mmapped at some point.

jadball:
Claude replies incoming comment-by-comment

ImageD11/sinograms/assemble_label.py
jadball: Claude:

Right, the `try/except` around the detector lookup was dropped. Restored in `write_scan_header`.

The other line that went was `title = hin[scan]["title"][()]`. That value was never used, and the `for check in ("title", ...)` loop above it already verifies `title` exists, so I left that one out. Say the word if you want it back.

ImageD11/sinograms/properties.py
jadball: Added:

```python
# TODO: the type should follow the input (int vs float). Casting to
# float64 here stops each pixel being truncated, but r is int64 below,
# so the sums and centroid moments are still truncated at the end.
```

Worth being explicit that this only half-fixes it: `r = np.empty((5, npks), np.int64)`, so `s_I`, `srI`, `scI` are still truncated once on assignment. The float cast only stops every individual pixel being truncated before the `bincount`. Making `r` follow the input dtype is the real fix and is a bigger change than this PR.

ImageD11/sinograms/properties.py
jadball: ## `pks_table.save` merge_options — "pks_table does not have a merge_options member in init. What is in here?"
and

## `rmem.merge_options = {...}` in main — "Belongs in init?"
Both fixed together. `merge_options` is now a keyword argument on `pks_table.__init__` (default `{}`), the keys are a module constant, and `pks_table.load` reads them back:

```python
MERGE_OPTION_KEYS = ("algorithm", "wtmax", "max_centroid_dist")
```

They are written as attributes of the `pks2d` group so you can look at a `pks.h5` and see how it was made. `None` is stored as the string `"None"` because h5py has no None.

`main` still assigns the attribute rather than passing it to the constructor, because `rmem` comes back from `goforit`/`pks_table_from_scan` and `main` is the only place that knows the options. Happy to thread it through the constructor instead if you prefer.

ImageD11/sinograms/properties.py
jadball: The intention was to merge only some of the overlapping pairs, but doing it inside `find_uniq` was the wrong place. Removed, see the reply to the review summary. `find_uniq` is now byte-identical to master.

ImageD11/sinograms/properties.py
jadball: Renamed to `pair_centroid_distance` and it now returns the distances rather than a boolean, so the name says what it computes and the caller writes the `if` itself:

```python
d = pair_centroid_distance(pkst.rc, pkst.pk_props)
pkst.rc = pkst.rc[:, d <= 1.6]
```

"gate" is gone from the code and the tests.

ImageD11/sinograms/properties.py
jadball: That falls out of the same shape now that the function returns a quantity per pair rather than a decision. `rc[2]` is already the shared pixel count and `pk_props[0]` is `s1`, so it is:

```python
frac = rc[2] / np.minimum(pk_props[0, rc[0]], pk_props[0, rc[1]])
pkst.rc = pkst.rc[:, frac > 0.25]
```

I have not added it, since you want the distance selections kept separate from the stored labels anyway — happy to put it in whichever PR it ends up belonging to.

ImageD11/sinograms/properties.py
jadball: Deleted. The two callers do the cut inline (4 lines each), so there is no wrapper hiding it and each caller can skip it.

ImageD11/sinograms/properties.py
jadball: Deleted. Both call sites now construct `SparseScan` directly:

```python
scan = ImageD11.sparseframe.SparseScan(hname, scans[i],
                                       names=SCAN_NAMES[options["algorithm"]])
```

where `SCAN_NAMES` is a dict of algorithm -> column names (and `ALGORITHMS = tuple(SCAN_NAMES)`).

To answer the question: yes, `'stored'` needs one extra pixel column, `labels`, plus a per-frame `nlabel` dataset alongside `nnz`. The `SparseScan` change is the one already in this PR — adding `"labels"` to `names` makes it read `labels` and `nlabel` and set `total_labels`, with the same convention as `lmlabel`/`cplabel(countall=False)`: labels run 1..nlabels[i] on frame i, 0 is background. It raises if `labels` is present without `nlabel`. I have documented that in the `SparseScan` docstring so the requirement lives in `sparseframe.py` rather than in `properties.py`.

ImageD11/sinograms/properties.py
jadball: Nobody gets it unless they ask: the default is `None` and the cut block is skipped entirely, so `lmlabel` on eiger data runs exactly the code path it runs on master. It is now distinct from `find_uniq` as described above.

ImageD11/sinograms/properties.py
jadball: Yes. `None` is the default and means "merge on any pixel overlap", i.e. current behaviour. `1.6` is what reproduces the old frelon KD-tree merge, and only the frelon notebooks set it.

ImageD11/sinograms/properties.py
jadball: Agreed on both counts. The shape of the code now admits other distance functions: anything that maps `rc` to a per-pair quantity can be cut on and the result handed to `find_uniq`, and `pair_centroid_distance` is written as an example of that rather than as the one blessed filter.

On splitting: they are separable in the code — the distance selection is `pair_centroid_distance`, the `max_centroid_dist` option, two 4-line cut blocks and the notebook parameters, and nothing else depends on it. I would rather keep them in one PR though, because they are only orthogonal in the abstract: the point of the stored labels is to move frelon onto the same merging code as eiger, and that route merges any pair of 2D peaks that share a pixel. On box-beam frelon data that is a behaviour change — peaks that merely touch get chained together — so stored labels landing on their own would be a regression for frelon users until the distance cut follows. `max_centroid_dist = 1.6` is what reproduces the existing merge, and it is what the frelon notebooks set. Splitting them means either merging a known regression first or holding the notebooks back until both are in.

Nothing is stopping the filters being generalised afterwards — the shape here is deliberately open for that, and I am happy to do the sparse-pixel-count filter and whatever Louise wants as a follow-up. If you still want them separate I will split them.

ImageD11/sinograms/assemble_label.py
jonwright: Does title hold fscan vs fscan2d vs f2scan ready for dataset? Or do we pick up all of fscan_parameters from the masterfile for that? Since you are in here, I suggest to harvest fscan_parameters if you are dropping title (or make it a todo in a new issue). In general, scan title belongs with the sparse pixels files. You need it and it is a tiny string.

ImageD11/sinograms/properties.py
jonwright: How does this work for a load and save without calling find_uniq? I seem to remember saving the pk2d and not being able to run the merge code with large data and small computers. In that case, the merge options doesn't make sense. Algorithm and wtmax are peak properties of all columns. 

Max centroid dist only applies to the glabel column and does not modify the others. I suspect you would want one list of pk2d for those peaks, and then be able to have several different glabel columns depending on which distance cut method you use. The distance cut is only a property for the glabel, not the others

ImageD11/sinograms/properties.py
jonwright: Historically we had the 1.6 pixel thing in the ancient (2005) peaksearch and got rid of it because some people hated it. With pixels as we have it now, the criteria to merge depends where you put the thresholds (high vs low) when looking at pixel overlap. But people could not accept anything else for deformed samples. For undeformed it doesn't matter what you do.

As it is now, the things are not yet set up for a few different cuts (and a dendogram, and the ability to split/merge peaks) later. You want to be able to cut a cf4d peak into two sub clusters, or not, and move the threshold later.

So on reflection, I suggest to leave glabel as it is "these have overlapping pixels" and then this distance thing becomes a new feature on top where we only need to do cuts inside each cf_4d peak. Does that make sense? The current clusters are the ones with the most links you can ever get. We don't change that meaning.

If there is a real scientific need to reproduce the frelon peaksearch output in this code, call it flabel and not glabel perhaps?

jonwright:
> Claude replies incoming comment-by-comment

Thanks - on reflection I think the cuts business is better done as splitting up the glabels - these have a current meaning that should not change (these peaks share pixels).

jonwright:
Having thought some more:

- in the past (#265) and future we also want to have a directed graph for localmax style labelling across frames

- the 1.6 distance cutoff should be using a different 'link' function. It is nonsense to compute a larger graph and then cut it down to be smaller, it should just use a different linkage function. You can also look at distance in dty/omega.

- the "stored" option for SparseScan needs some extra argument (column label to be loaded). Same problem for lmlabel, there is a hidden "smooth" option.

So there is a bunch of refactoring waiting in #265 as well as cleanups like #602, #479 (links missing across cut?), and #469.

These changes add a "new feature" inside an already complex code. It would make more sense for what is wanted:

- load the SparseScans
- load or compute labels for them (stored vs labelling)
- compute 2d moments
- compute linkage/overlap graph and copy into rc from the moments
- derive flabels (or mlabels, or some other new name)

Then you skip and O(npixels) step and only do the O(npeaks) instead. This is a different series function calls. 

If you save centroids when you are doing the labelling, there is no need to load the SparseScans again. This seems to clear a path towards removing the shm and cleaning the whole thing up. Get the 1.6 pixel version working without shm, and then add the pixel overlaps (and directed graphs) as alternatives top of that, rather than the other way around.

jonwright:
OK - I will continue to work on a refactor to change linkage functions and take the saving and loading of labels from here. As it is, we can't make the frelon code equivalent to the sparse by this route (two adjacent single pixels are merged by frelon and not by sparse). The first incoming is a cleanup of SAFE that was wrongly set.

PR #638, opened 2026-09-17
Frelon sparse pixels

jonwright:
Notes below are from the LLM. Code needs to be double checked read by a human before hitting master.

## What this branch is

The first piece of replacing PR #633 (@jadball, "Unify frelon and eiger peak merging
via sparse pixel files"). It lets the frelon segmenter write a sparse pixel file that
the existing readers understand, and lets `properties` use labels from that file.

**It changes nothing for anyone installing from master.** Everything added is opt-in.
Verified: default algorithm is still `lmlabel`, `SCAN_NAMES['lmlabel']` matches the old
`SparseScan` default columns, integer pixels still give `int64`, `segment_dataset` is
untouched and unwarned, and the notebooks are byte-identical to master.

## Done

| commit | what |
| --- | --- |
| `0744540` | `assemble_label.write_scan_header` split out of `harvest_masterfile`, and the scan title written into the sparse file instead of read and dropped |
| `70bddda` | `worker.peaksearch_sparse`, `pps_sparse`, `frames_by_scan`, `segment_dataset_to_sparse`, `_write_sparse` |
| `1241836` | `moment_dtype` so `pk_props` follows the pixel dtype; `algorithm='stored'`; `SparseScan` reads `labels`/`nlabel` |
| `4867fce` | gitignore for test run artifacts |
| `805cacb` | notebook reverted, so the branch is inert for users |
| (next) | the sparse route shown in the notebook as a commented cell, default untouched |

`test/test_frelon_sparse.py`, 12 tests, including the whole route: segment to sparse,
then `properties.main(algorithm='stored')`. Full suite 252 passed.

## Not done

- [ ] **Notebook wiring.** The S3DXRD frelon notebook shows the sparse route in a
      commented cell, with a note saying it merges differently. The default is still
      `segment_dataset` and nothing live changed. Making it the default needs the merge
      question settled first.
- [ ] **TDXRD `0_segment_frelon.ipynb`.** Produces `cf_3d` and writes `col3dfile`; the
      properties route produces the 4D output. Whether `get_cf_4d` substitutes for
      `cf_3d` on a box-beam dataset with no dty scanning is unverified. PR #633 changed
      it; I did not.
- [ ] **No way to reproduce the old frelon peaks.** This is the blocker for the
      notebooks. See the incomparability note below.
- [ ] `frm` is not in `columnfile.INTS`, so it is already a float column. Unrelated but
      noticed.

## Decisions already made, do not relitigate

- **The two merge rules are incomparable.** Legacy frelon merges 2D peaks whose
  *centroids* are within 1.6 px (`do3dmerge`, KD-tree, no pixel test at all). The sparse
  route merges peaks that *share a detector pixel* (`pairrow`/`pairscans` →
  `sparse_overlaps`). Neither contains the other: two 1-pixel peaks 1.0 px apart share
  no pixel; two long peaks sharing one pixel can have centroids 10 px apart. So
  **cutting the overlap graph cannot reproduce the old merge**, which is why
  `max_centroid_dist` and `pair_centroid_distance` from PR #633 are not here.
- **`glabel` keeps its meaning**, "these peaks share pixels". A different criterion gets
  a different column name, never a redefinition.
- **`concurrent.futures`, not `tqdm.contrib.concurrent.process_map`.** `process_map`
  appears only in `frelon_peaksearch.py`; `concurrent.futures` is used in seven modules.
  PR #633 was cleaning up an outlier. I reverted that once by mistake.
- **float32 intensity is correct**, not a choice: `worker.cor` is float32 after
  background subtraction (`frelon_peaksearch.py:303`).
- **Do not set `g.attrs['itype']` to the intensity dtype.** It means two different
  things already: the *index* dtype in `to_hdf_group`/`from_hdf_group`, the *intensity*
  dtype in `write_scan_header`/`lima_segmenter`. Pinned by a test.
- **`SparseScan` casts intensity to float32 on read** and keeps the stored dtype in
  `intensity_input_dtype`. `moment_dtype` must be given the *stored* dtype, or the eiger
  route silently becomes float64 and every existing result changes.
- **`signal.min() <= 0`** is identical to `< 1` for integers and correct for floats.

## Dropped from PR #633 and staying dropped

`max_centroid_dist`, `pair_centroid_distance`, the `rc` cut, `MERGE_OPTION_KEYS` /
`merge_options`, the `segment_dataset` deprecation, and the old-vs-new merge assertions
in `test_sparse_merge.py` (they assert an equivalence that cannot hold, and pass only
because the synthetic data contains neither divergent case).

## How to run things here

```
module load jupyter-slurm
PY=/cvmfs/hpc.esrf.fr/software/packages/linux/x86_64/jupyter-slurm/2025.04.6/envs/jupyter-slurm/bin/python3
$PY setup.py build_ext --inplace          # cImageD11 is not built in this checkout
PYTHONPATH=$PWD $PY -m pytest test/test_frelon_sparse.py -q
PYTHONPATH=$PWD env -C test $PY -m pytest -q --ignore=test_fetch_data
```

Do not run the papermill notebooks here. Real data at
`/data/id11/jon/blc14570/id11/20230425/PROCESSED_DATA` is **read only**, do not write there.

## Next, in order

1. This branch reviewed and merged (or parked, we will branch off of it)
2. Omega wrapping and which frames are neighbours.
3. The linkage function: overlaps vs centroids vs others.
4. How to iterate over a sinogram. This is the only one that touches the memory wall
   (`pk_props` 17 GB and `rc` >= 9.6 GB at 4.24e8 peaks; 120 GB and >= 68 GB at 3e9).

One PR at a time, reviewed, then merged. They all touch `properties.py` and will
conflict otherwise.


jonwright:
Leaving this unmerged for now. as we are blocking on point 2 above.

PR #642, opened 2026-09-20
Adding a peak property called max intensity in  columnfile to identify saturated peaks ( can be useful to improve strain refinement )

aditya-shukl:
Saturated peaks on eiger are being replaced by a user defined value of wtmax in sinograms.properties. It works well but information about peak saturation is lost downstream. Saturated peaks can  affect COM of a peak and can affect strain calculations downstream . So,  I added a peak property that can be called as columnfile.IMax_int. I use this to filter out the saturated peaks before strain refinement.


jonwright:
Thanks @aditya-shukl . I can merge as it is now, but I will want to change the behavior. It is spending RAM for a column that we can live without. A few options:

- return the pk2d with these peaks already removed (easiest)
- make a matching peaks file with only saturated peaks in it only, figure out the labels+glabels of these two peaks files
- add a list of lots of different 'flags', like "touches_mask", "exceeds_wtmax", "on_first_last", etc. Then fill these into a bitmask.
- make is easier to get from peaks back to pixels to do this filter by getting the pixel list for each peak and filtering on max

If you use `wtmax = exposure_time_in_seconds * 10^7` then the positions are often quite good anyway.

PR #643, opened 2026-09-21
pbp_refine: NaN, not 0.0, for peaks with no diffraction origin

aditya-shukl:
In get peak origin, all origins are initialized with a value zero.  I changed it to Nan so that these peaks are not used in refinement. ( maybe zero was also okay? ). But, Nan placeholder value can also be useful to assign to peaks where get origin fails for example in donut shaped grains. I am working on a pull request for that.


jadball:
I understand the concept here, but I'm a little concerned about a few points:

- this changes the refinement process quite a lot. in the past, peaks with no assigned xpos_refined were still considered part of the refinement. now, they are excluded. Perhaps @AxelHenningsson can weigh in?
- I don't know how the refiner will handle NaN in xpos_refined. will it fail gracefully? I'm concerned it will crash or silently skip parts of the sample during refinement if it encounters them. I don't see NaNs being handled inside the refiner currently aside from checking how many there are.
- I'm not sure how the columnfile class will handle NaN, especially saving/loading HDF5. needs testing I think

What about an alternative:

- after doing get_origins, check to see which peaks have a non-zero xpos_refined
- add a user parameter to the refinement telling it whether or not to use these zero peaks

jadball:
(if you push more commits to the `pbp-origin-validity` branch it will update this pull request)

and - thanks for contributing!

aditya-shukl:
Hi James. Thanks for the suggestions. I used refine.to_h5(save_peaks=True) to save the peaks file with xpos refined nans . I also loaded with refine = PBPRefine.from_h5(refmanpath) to check refined.icolf. had nan values. I also compared strain refinement before and after the change on my Ni  data, and could not spot a difference in the maps. Happy to test more things  if you have suggestions. Physically what does xpos refined =0 mean? Is it  not possible for a real peak to have xpos refined as zero?

jonwright:
Did you look at computing number of voxels contributing to a peak? Then n==0 is your nan and you can remove them. It gives you Icalc for free to close the loop on iterative weights.

Nan tends to propagate, so it is different to unknown, weight==0.

PR #644, opened 2026-09-21
Donut grain peak origin

aditya-shukl:
In case of donut shaped grains, twin grains ( grain, twin grain , grain again) or many grains along the xray path , as in the case shown below, some peaks can get wrong weighted origin values. There needs to be a check after origin computation that decides if the peak origin actually sits on a grain that the peak likes. is this problem taken care of during refinement? if not ,  these peaks should be given a nan ( implemented in this pr) or a zero as xpos refined . See the image below that descibes the problem . The image was generated with an hkl tol of 0.02 for the peak . I checked the changed code on my Ni data and it gives me the same strain map  image


jonwright:
It sounds like you want to always remove all peaks with overlaps above. The code reads as if it will semi-randomly suppress peaks from concave or disconnected regions. If you see a dendrite there can be a lot of disconnected regions, you might want to keep them, or not, and depending on tolerance and hkl_tol is it not obvious if are removed. If you put a large grain next to small twin, the small twin still has an overlap problem and looks like it still gets bad strains from some overlapping peaks.

I suspect the same comment to the other issue applies... Do you want to assign 'grain_id' to each disconnected region of your map? Then in the forward projection to compute lx, count how many different grains contribute to a peak? In the other issue you wanted `(contributions > 0)`, in this one you want to ensure all contributions are from the same grain_id. 

So far as I am aware, we don't use NAN for a mask in ImageD11. You get into trouble with floating point exception handling so it is better to have a proper mask.

Maybe a GUI/picture would help. You click on a pixel in the strain map and it could light up the lx contributions along each ray for each peak used in the fit...

aditya-shukl:
yeah,  i agree removing all such peaks from disconnected regions may not be the best idea . I will close this pr and maybe open it as an issue

PR #645, opened 2026-09-21
Refactor friedel

jonwright:


jbjacob94:
I had a look at the code in friedel_pairs.py and match_friedel_pairs.py and did some quick tests with an older dataset. A few thoughts and comments below:

**1 - some cleaning is needed**
Lot of new additions, the code now looks very convoluted and difficult to read and maintain (e.g. function names and class attributes still use legacy names, quite confusing).  Some ad hoc pairing function used in different notebooks have just been moved at the end of the file. This  should be removed and notebooks should be updated to use one single pairing method. Also,  I think fit_y0 function in friedel_pair.py should go elsewhere. 

There is a core pair matching function _run_pairing(cf, i1,i2, pair_type, **options), which takes two subsets of cf (given by i1,i2 indices arrays) as input and returns paired peaks in these two subsets as output. This should run for any type of pairs and is independent of how the subsets are defined: for global pair matching as done in the notebooks  i1 and i2 are just defined by splitting cf in two half (e.g. by +/- eta)
I think it would be better to use it in the notebooks instead of the ad-hoc functions. 

**2 - Peaksubset needs to be updated** 
Deepseek has not understood how pairs of subsets are built, and therefore cannot match properly the new "horizontal" pairs because it tries to find them in unrelated subsets. I have the feeling this part could be simplified:

In the legacy version:  three ways of subsetting cf: dty bins ('scans'), omega_dty bins ('frames') and eta_bins. Then, pairs of subset are built according to where the match are expected to occur for the different pairs: 
- omega (vertical) pairs match by scans:  mirror scans at dty = (y0 + dy), (y0-dy)
- omega pairs match by frames:  mirror frames at  (y0+dy,omega) and (y0-dy,omega+180)
- eta (diagonal) pairs match by eta bins: opposite bins at eta, eta+180 

new suggestion: all subsets could be defined in terms of eta bins: (eta,eta+180) for diagonal pairs, (eta,180-eta) for vertical pairs and (eta, -eta) for horizontal pairs. Also more logical with new pair names. I would like to keep the legacy scans/frames subset definition somewhere though, as they are more closely tied to the acquisition (cf is sorted in sinogram order and subsets can be defined to correspond to a single scan or frame). 

** 3 - What to do with "match_quadruplets"?
In the legacy version, these were found by finding blocks of 4 peaks which mutually satisfy both 'vertical' and 'diagonal' relationships

eta pair 0     eta pair 1
omega pair 0:   peak p00  ——  peak p01
                                       |               |
 omega pair 1:   peak p10  ——  peak p11

With the new pairing it would be more consistent to define the 4-blocks using the vertical, horizontal and diagonal pair of each peak. 

**4- match_friedel_pairs.py**
Standalone pipeline script to run pairing in large peakfiles and send jobs to slurm. It still uses the legacy names. This should be updated with new pair names and include an option for matching horizontal pairs. 

**5 - notebooks**
I haven't checked in detail, but probably need to be updated to use the same core pairing function. 
match_friedel_pairs.ipynb: demo notebook to explain how pairing works, not used in any pipeline. needs to be updated to add new pair names and but this is not a priority

jonwright:
Thanks! I will call by later to talk about it. The new functions pair on g-vectors and the existing ones used angles. For an inclined axis it seems to get complicated.

jonwright:
Optimistic that this can be cleaned up, but there are too many changes for 1 PR: switching to horiz/vert/diag, conceptually different problems for cf_4d vs cf_2d vs box beam. Notebooks could see how to call into `_run_pairing` without the other changes.
End-to-end type test cases are needed to validate/compare. The notebook code is reporting all matches, the original code is selecting the best match in each case, and looking at intensities. Chunking is needed, and a common issue for voxel mask as well as feeding frames to peaksearch.

